PR #28, opened 2018-04-10
my second commit

rachelz11:


rachelz11:
thanks Harry. It is indeed a much better way to do this task. i will update my answer.

PR #29, opened 2018-04-11
update break_me with function

rachelz11:
I changed my break_me with function as harry suggested.


WesWang/00-README.md
```diff
@@ -1,3 +0,0 @@
-Hi, my name is Wesley Wang.
-Feel free to call me Wes.
-I look forward to mastering Python after this class.
```
hosungs: Thanks for removing this! I should have reviewed earlier PRs (not just blindly merging them) and pointed this out.

students/Ruohan/lesson01/break_me.py
```diff
@@ -1,22 +1,25 @@
...
+    print ('NameError: ')
+    print (y)
 
 #TypeError
-#print ('TypeError: ')
-#x = 2
-#y = 'hello'
-#print(x+y)
+def give_TypeError():
+....print ('TypeError: ')
```
hosungs: Do I see actual dots here at the beginning of this line? Wonder if it didn't give a SyntaxError in the first place?

PR #30, opened 2018-04-13
Meet Shah: Lesson02

meet29in:


students/meet_shah/lesson02/print_grid.py
```diff
@@ -0,0 +1,54 @@
+def print_grid(x):
```
HarryMaher: Looks great!

students/meet_shah/lesson02/fizzbuzz.py
```diff
@@ -0,0 +1,11 @@
+def fizzbuzz():
```
HarryMaher: Nice! Good idea.

students/meet_shah/lesson02/print_grid.py
```diff
@@ -0,0 +1,54 @@
...
+    topstart = "+ "
+    topmid   = "- "
+    topend   = "+"
+
+    midstart = "| "
+    midmid   = "  "
+    midend   = "|"
+
+    repeat = int(x/2)
+    topline = topstart + topmid*repeat + topstart + topmid*repeat + topend
```
HarryMaher: You can multiply help make it more DRY and less typing. Eg:
```
topline = (topstart + topmid * repeat) * 2 + topend
```

students/meet_shah/lesson02/print_grid.py
```diff
@@ -0,0 +1,54 @@
...
+        for i in range(0,repeat):
+            print(midline)
+        print(topline)
+ 
+def help_print_line(startpattern, endpattern, repeat):
+    for i in range(0,repeat):
+        print(startpattern, end='')
+    print(endpattern)
+
+def print_grid2(box_repeat,grid_repeat):
```
HarryMaher: Awesome!

students/meet_shah/lesson02/print_grid.py
```diff
@@ -0,0 +1,54 @@
...
+    midline = midstart + midmid*repeat + midstart + midmid*repeat + midend
+    
+    print(topline)
+    for i in range(0,2):
+        for i in range(0,repeat):
+            print(midline)
+        print(topline)
+ 
+def help_print_line(startpattern, endpattern, repeat):
+    for i in range(0,repeat):
```
HarryMaher: I like that you used a helper function to break up the code!
You can simplify this a bit by getting rid of the loop here and just using string multiplication. 
``` 
ret = startpattern * repeat + endpattern
print(ret)
```

students/meet_shah/lesson02/print_grid.py
meet29in: ah sweet! totally missed that.

meet29in:
I pushed part3 late Sunday night. It's not past due yet but I was curious if it looked good or if I need to the code here.

students/meet_shah/lesson02/series.py
```diff
@@ -0,0 +1,32 @@
+
+def fib(n):
```
HarryMaher: Cool, looks good. We also wanted a sum_series function, but this look like you've good the idea. Check out the solution we did in class if that was causing you issues.

meet29in: yep missed the sum_series :(

PR #31, opened 2018-04-13
OSiddiquee PR for lesson02

osiddiquee:
I have finished the 3 exercises. This is my PR for these exercises.

### DIFF
--- a/students/Osiddiquee/lesson02/FizzBuzz.py
+++ b/students/Osiddiquee/lesson02/FizzBuzz.py
@@ -1,0 +1,9 @@
+for n in range (1,100):
+    if n%15 == 0:
+        print('FizzBuzz')
+    elif n%5 == 0:
+        print('Buzz')
+    elif n%3 == 0:
+        print('Fizz')
+    else:
+        print(n)
--- a/students/Osiddiquee/lesson02/grid.py
+++ b/students/Osiddiquee/lesson02/grid.py
@@ -1,0 +1,15 @@
+border = '+' + 4 * ' -' + ' +' + 4 * ' -' + ' +'
+interior = '|' + 9 * ' ' + '|' + 9 * ' ' + '|'
+
+
+print(border)
+print(interior)
+print(interior)
+print(interior)
+print(interior)
+print(border)
+print(interior)
+print(interior)
+print(interior)
+print(interior)
+print(border)
--- a/students/Osiddiquee/lesson02/print_grid.py
+++ b/students/Osiddiquee/lesson02/print_grid.py
@@ -1,0 +1,13 @@
+def print_grid(n):
+    border = '+' + n//2*' -' + ' +' + n//2* ' -' + ' +'
+    if n % 2 == 0:
+        interior = '|' + (n+1)*' ' + '|' +  (n+1)*' ' + '|'
+    else:
+        interior = '|' + n*' ' + '|' +  n*' ' + '|'
+    print(border)
+    for x in range(0,n//2):
+        print(interior)
+    print(border)
+    for x in range(0,n//2):
+        print(interior)
+    print(border)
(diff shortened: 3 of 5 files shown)


students/Osiddiquee/lesson02/FizzBuzz.py
```diff
@@ -0,0 +1,9 @@
+for n in range (1,100):
+    if n%15 == 0:
```
HarryMaher: Nice! I like the clever least common multiple idea.

A couple of notes:
- range(1, 100) returns to 99, and I think the assignment wanted to 100. The second argument of range should be one higher than you want.
- Here's what the spacing should be, stylistically, according to PEP8:
```
for n in range(1, 101):
    if n % 15 == 0:
#etc. etc.
```

students/Osiddiquee/lesson02/grid.py
```diff
@@ -0,0 +1,15 @@
+border = '+' + 4 * ' -' + ' +' + 4 * ' -' + ' +'
+interior = '|' + 9 * ' ' + '|' + 9 * ' ' + '|'
+
+
+print(border)
+print(interior)
+print(interior)
+print(interior)
+print(interior)
+print(border)
```
HarryMaher: Not the most DRY code, but you already have the loops idea in your other submissions for this, so I think you got the idea.

students/Osiddiquee/lesson02/print_grid.py
```diff
@@ -0,0 +1,13 @@
+def print_grid(n):
```
HarryMaher: Great!

students/Osiddiquee/lesson02/print_grid.py
```diff
@@ -0,0 +1,13 @@
...
+    if n % 2 == 0:
+        interior = '|' + (n+1)*' ' + '|' +  (n+1)*' ' + '|'
+    else:
+        interior = '|' + n*' ' + '|' +  n*' ' + '|'
+    print(border)
+    for x in range(0,n//2):
+        print(interior)
+    print(border)
+    for x in range(0,n//2):
+        print(interior)
```
HarryMaher: Instead of repeating the above code, you can put the loops in a loop

students/Osiddiquee/lesson02/print_grid2.py
```diff
@@ -0,0 +1,16 @@
+def print_grid2(x,y):
```
HarryMaher: Good job! 
A few small things:
- x and y aren't the most descriptive parameter names
- the first two loops of loops can be replaced with using string multiplication
Overall good work!

students/Osiddiquee/lesson02/print_grid2.py
```diff
@@ -0,0 +1,16 @@
...
+    border = '+'
+    interior = '|'
+    for n in range (0,x):
+        for n in range (0,y):
+            border += ' -'
+        border += ' +'
+    for n in range (0,x):
+        for n in range (0,y*2):
+            interior += ' '
```
HarryMaher: Instead of  range(0, y * 2) and increasing the number of loops, you can make interior += two spaces instead of one. 
Loops can be expensive computationally. Instead of loops you can do something like: 
```
border = "+" + (" -" * y + " " + "+") * x
```

students/Osiddiquee/lesson02/series.py
```diff
@@ -0,0 +1,37 @@
...
+def lucas(n):
+    if n == 0:
+        return 2
+    elif n == 1:
+        return 1
+    else:
+        return lucas(n-1) + lucas(n-2)
+
+#nth paramater of generalized series function
+def sum_series(n, first = 0, second = 1):
```
HarryMaher: Good!

PR #34, opened 2018-04-15
Print Grid Exercise Completed --kristianjf

kristianjf:
Completed print grid exercise parts 1-3.


students/kristianjf/lesson02/fizzbuzz.py
```diff
@@ -0,0 +1,17 @@
...
+But for multiples of three print “Fizz” instead of the number.
+For the multiples of five print “Buzz” instead of the number.
+For numbers which are multiples of both three and five print “FizzBuzz” instead.
+'''
+
+for x in range(1,101):
+    if (x % 3 == 0) and (x % 5 == 0):
+        print('FizzBuzz')
+    if (x % 3 == 0):
+        print('Fizz')
```
HarryMaher: Awesome!

students/kristianjf/lesson02/print_grid.py
```diff
@@ -0,0 +1,58 @@
...
+|         |         |
++ - - - - + - - - - +
+'''
+a = '+ '
+b = '- ' * 4
+c = '| '
+d = '  ' * 4
+x = str('{}{}{}{}{}\n').format(a, b, a, b, a)
+y = str('{}{}{}{}{}\n').format(c, d, c, d, c)
+print('Part 1: Write a function that draws a grid\n'+x+y*4+x+y*4+x)
```
HarryMaher: Whoa I really like that you used .format() and "\n" and completely eliminated the need for looping. Very clever!

students/kristianjf/lesson02/print_grid.py
```diff
@@ -0,0 +1,58 @@
...
+'''
+Part 2: Make it a function
+Write a function print_grid(n) that takes one integer argument and prints a grid just like before, BUT the size of the grid is given by the argument.
+'''
+def print_grid(x):
```
HarryMaher: Great!

students/kristianjf/lesson02/print_grid.py
```diff
@@ -0,0 +1,58 @@
...
+    n = x // 2
+    a = '+ '
+    b = '- ' * n
+    c = '| '
+    d = '  ' * n
+    x = str('{}{}{}{}{}\n').format(a, b, a, b, a)
+    y = str('{}{}{}{}{}\n').format(c, d, c, d, c)
+    print('Part 2: Write a function print_grid(n) that takes one integer argument and prints a grid just like before, BUT the size of the grid is given by the argument.\n'+x+y*n+x+y*n+x)
+
```
HarryMaher: Small quibble: PEP 8 says that code shouldn't be more than 79 chars in a line. So just be aware that ideally this would be shorter, and that you can break code up with a "\". But, this is an awesome solution!

students/kristianjf/lesson02/print_grid.py
```diff
@@ -0,0 +1,58 @@
...
+print_grid(15)
+'''
+Part 3: Write a function that draws a similar grid with a specified number of rows and columns, and with each cell a given size.
+
+Note = Exercise is to print length * width based on first input and number of lines within each square by second input
+'''
+def print_grid2(rows_columns=1,size=1):
```
HarryMaher: Really good work

students/kristianjf/lesson02/series.py
```diff
@@ -0,0 +1,64 @@
...
+    for x in range(1,n):
+        n3 = n1 + n2
+        luc_list.append(n3)
+        n1 = luc_list[-2]
+        n2 = luc_list[-1]
+    return luc_list, luc_list[n]
+
+print('Lucas List: {} \nLucas Value: {}'.format(Lucas(7)[0],Lucas(7)[1]))
+
+def sum_series(n, n1=0, n2=1):
```
HarryMaher: Good, I like the logic. It's clean and easy to follow. 
A few comments: 
For all these, we really only needed the nth value of the series, so storing all the values in a sum_list can get computationally costly if we ask for a really high number. Another option would be to store the total sum and only hold onto n1 and n2 as variables. Something kind of like this:
```
def fib2():
    a = 0
    b = 1
    so_far = 0
    for i in range(10):
        so_far += a
        a, b = b, a + b
    print(so_far)
fib2()
```

PR #36, opened 2018-04-15
Andy Kwon - Lesson02 PR (Grid Printer, FizzBuzz and Series)

KwonAndy:
Committed and pushed 2 Grid Printers for our homework. The first one (gridprinter.py) takes in 1 parameter. The second file (gridprinter2.py) takes in 2 different parameters to formulate a grid. 

Thank you,

Andy


students/andykwon/lesson02/gridprinter.py
```diff
@@ -0,0 +1,14 @@
+def print_grid(num):
```
HarryMaher: Looks good!

In the future you would probably want to store all of the repeated parts in a variable name.  If your boss was like, "oh no we wanted ampersands and not plus signs" it would be tedious to go through and copy and paste new code and this could lead to errors. I'd recommend storing these and then printing them instead, next time:
```
top_line = '+' + ' - ' * int(num/2) + ' + ' + ' - ' * int(num/2) + '+'
middle_line = '| ' + ' ' * int(num/2)*3 + '|' + ' ' * int(num/2)*3 + ' |'
```

students/andykwon/lesson02/gridprinter2.py
```diff
@@ -0,0 +1,22 @@
+def print_grid2(num1, num2):
```
HarryMaher: Nice grid!
Some feedback:
- If you like the extra newlines for readability, that's fine, but note that that amount of space isn't necessary. Whatever you prefer
- num1 and num2 aren't the best parameter names. Something a little more descriptive would be ideal. It'd help with readability.
- For both gridprinter2 and gridprinter, you have twice as many spaces between the dashes as specified. Not a huge issue, just a heads up. 
- You can make this a little more simple if you use string multiplication instead of for loops in some of these cases. Loops can be computationally expensive, so for example instead of:
```
for j in range (num1):
    print(end='+' + ' - ' * num2)
```
you can do something like:
```
print(("+" + " - " * num2) * num1 + "+")
```

KwonAndy:
Thank you for the feedback Harry! Yes, some of the comments you mentioned such as the parameter names and also the the storing of repeated parts came to mind afterwards. And the computationally heaviness of loops are something I had not considered. I'll definitely keep that in mind going forward. 

Thanks again!

Andy

PR #39, opened 2018-04-16
Lesson 2 Assignments - WesWang

WesWang93:
FizzBuzz, Fibonacci/Lucas Series, Grid Printer

### DIFF
--- a/students/WesWang/lesson2/fibonacci_lucas.py
+++ b/students/WesWang/lesson2/fibonacci_lucas.py
@@ -1,0 +1,29 @@
+#=========================================================================================
+#Name: Fibonacci Series & Lucas Number
+#Author: Wesley Wang
+#Date: 4/12/2018
+#=========================================================================================
+
+def sum_series(n0, n1, nth):
+    prev = n0
+    current = n1
+    if nth > 2:
+        for i in range(nth-2):
+            prev, current = current, prev + current
+    if nth == 1:
+        return prev
+    elif nth <= 0:
+        return "Please enter a non-zero, positive number!"
+    else:
+        return current
+
+def fibonacci(nth):
+    return sum_series(0, 1, nth)
+
+def lucas(nth):
+    return sum_series(2, 1, nth)
+
+
+print(fibonacci(5))
+print(lucas(5))
+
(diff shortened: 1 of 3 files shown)


students/WesWang/lesson2/fizz_buzz.py
```diff
@@ -0,0 +1,26 @@
...
+#=========================================================================================
+
+"""
+This function return interger 1 ~ 100
+If the number is divisible by 3, return "Fizz"
+If the number is divisible by 5, return "Buzz"
+If the number is divisible by 3 and 5, reutnr "FizzBuzz"
+"""
+def fizz_buzz():
+    result = ""
```
HarryMaher: Cool. I like it

students/WesWang/lesson2/gridprinter.py
```diff
@@ -0,0 +1,29 @@
...
+#=========================================================================================
+
+"""
+Used by print_grid to draw horizontal line(floor)
+:param: Number row/col, Size of grid
+:return: Floor with specified row/col count and size
+"""
+def draw_floor(rowcol, size):
+  for x in range(rowcol):
+      print("+" + "-"*size, end='')
```
HarryMaher: Cool, I like your helper function! Another option would be to use string multiplication instead of using a loop here, but this works too! (Too many loops within loops can get computationally expensive especially as the grid size and number of rows increases)

students/WesWang/lesson2/fibonacci_lucas.py
```diff
@@ -0,0 +1,36 @@
...
+    current = n1
+    if nth > 2:
+        for i in range(nth-2):
+            prev, current = current, prev + current
+    if nth == 1:
+        return prev
+    elif nth <= 0:
+        return "Please enter a non-zero, positive number!"
+    else:
+        return current
```
HarryMaher: Nice work!

students/WesWang/lesson2/fibonacci_lucas.py
```diff
@@ -0,0 +1,36 @@
...
+
+#Pass in 0 and 1 as first two values and nth to sum_series to calculate value at nth position
+def fibonacci(nth):
+    return sum_series(0, 1, nth)
```
HarryMaher: Good job on using the sum_series as a helper function

PR #41, opened 2018-04-16
added all files from assignment 01

aravichander:


students/aravichander/lesson02/PrintGrid.py
```diff
@@ -0,0 +1,26 @@
+#This definitely did not work the way I wanted it to and I ran out of time before I could fix it!
```
HarryMaher: Yeah these grids didn't work out this time, but looks like you gave it a shot, so no worries. Please check out the code in the "Solutions" folder to get an understanding for how you can approach this problem. If that code doesn't make sense, sometimes the best way to learn is to copy it into your editor and put a bunch of print() statements to see what's happening at certain points.

Let me know if you have any questions about the solution we posted, and come to office hours or email if you want more help with anything!

PR #42, opened 2018-04-17
lesson 2 commit

DwightJay:
Submitting lesson 2.


students/jayjohnson/lesson01/break_me.py
```diff
@@ -0,0 +1,37 @@
...
+'''
+
+# error functions
+def nameError_fuction():
+    a = 5
+    del a
+    a
+
+def typeError_fuction():
+    type = 'abc'
```
HarryMaher: Since the point of this exercise is getting errors, reassigning a builtin function is a good way to do that I suppose, so good work :). An example of a more common type error could be like:
```
def typerr_fun():
    return "35"/5
```

students/jayjohnson/lesson01/break_me.py
```diff
@@ -0,0 +1,37 @@
...
+    type('efg')
+
+def syntaxError_fuction():
+    #del = 'error please'
+
+def AttributeError_fuction():
+    pass
```
HarryMaher: A good way to get an attribute error is by trying to access an attribute of something that doesn't exist. (We'll learn more about this later).
So for example:
```
#This throws an attribute error:
"this is a string".made_up_attribute
#this doesn't:
"blah".isalpha()
```
And, if you want a list of attributes for an object, use dir()!
```
dir("string")
```

students/jayjohnson/lesson02/Fizz_Buzz.py
```diff
@@ -0,0 +1,28 @@
...
+x For numbers which are multiples of both three and five print “FizzBuzz” instead.
+'''
+# loop logic
+guess = False
+
+# count set to 100
+test_num = 100
+
+# loop
+while guess == False:
```
HarryMaher: An interesting fizzbuzz! The logic works. 
A few things to think about:
- This goes in the opposite order from usual. How would you do it in the traditional 1-100 order?
- guess is False, but then you never reassign it, so it's always False and you just break from the loop with break. So you can take out the guess and just start with:
```
while True
```

students/jayjohnson/lesson02/Grid_Printer.py
```diff
@@ -0,0 +1,151 @@
...
+def print_grid_simple_function(n):
+    print('+ ' + '- ' * n  + '+ ' + '- ' * n  + '+')
+    print('|' + '  ' * n + ' |'  + '  ' * n + ' |')
+    print('|' + '  ' * n + ' |'  + '  ' * n + ' |')
+    print('|' + '  ' * n + ' |'  + '  ' * n + ' |')
+    print('|' + '  ' * n + ' |'  + '  ' * n + ' |')
+    print('+ ' + '- ' * n  + '+ ' + '- ' * n  + '+')
+    print('|' + '  ' * n + ' |'  + '  ' * n + ' |')
+    print('|' + '  ' * n + ' |'  + '  ' * n + ' |')
+    print('|' + '  ' * n + ' |'  + '  ' * n + ' |')
```
HarryMaher: This works. It's not DRY, so you'd want to look into loops to get this looking less repetitive, but alright!

students/jayjohnson/lesson02/Grid_Printer.py
```diff
@@ -0,0 +1,151 @@
...
+
+def print_grid(n):
+    grid_count_upper = math.floor(n/2)
+    grid_count_lower = math.floor(n/2)
+    print('+ ' + '- '*math.floor(n/2) + '+ ' + '- '*math.floor(n/2) + '+'  )
+    for i in range( grid_count_upper ):
+        print('| ' + '  '*math.floor(n/2) + '| ' + '  '*math.floor(n/2) + '|'  )
+    print('+ ' + '- '*math.floor(n/2) + '+ ' + '- '*math.floor(n/2) + '+'  )
+    for i in range( grid_count_lower):
+        print('| ' + '  '*math.floor(n/2) + '| ' + '  '*math.floor(n/2) + '|'  )
```
HarryMaher: There are still some repeated lines, so putting everything here in a loop would help with that. Also assigning repeated strings to variables helps with readability and makes it easier to edit later on. What if, for example, we changed our minds and wanted "&" instead of "+". This would require some copy pasting which could lead to introducing errors.

students/jayjohnson/lesson02/series.py
```diff
@@ -0,0 +1,96 @@
...
+def lucas(n):
+    if n == 0:
+        return 2
+    if n == 1:
+        return 1
+    return lucas(n - 1) + lucas(n - 2)
+
+def sum_series(n,x=0,y=1):
+    if x == 0 and y == 1:
+        return fibonacci(n) + lucas(n)
```
HarryMaher: Hopefully you were able to follow along in class, and if not we can talk about it later! No worries.

students/jayjohnson/lesson02/series.py
```diff
@@ -0,0 +1,96 @@
...
+print("This is each series counted out to the 7th term")
+print("Fibonacci series:")
+print(fibonacci(0))
+print(fibonacci(1))
+print(fibonacci(2))
+print(fibonacci(3))
+print(fibonacci(4))
+print(fibonacci(5))
+print(fibonacci(6))
+print(fibonacci(7))
```
HarryMaher: Cool! Later we'll learn about assert statements and tests, but this is one way to test the outcome for sure

students/jayjohnson/lesson02/Grid_Printer.py
```diff
@@ -0,0 +1,151 @@
...
+|       |       |       |       |       |
++ - - - + - - - + - - - + - - - + - - - +
+'''
+
+def print_grid2(c,n):
+    print('+' + " - "*n + '+')
+    for i in range( n ):
+        print('|' + "   "*n + '|')
+    print('+' + " - "*n + '+')
+    #cant figure out how to print multiple boxes
```
HarryMaher: Check out the solutions folder and see if that makes sense to you. Let me know if you need more of an explanation. I find it helpful to copy code into my editor, and put in a ton of print() calls to see how things change as the program progresses. Also, commenting out code can help.

students/jayjohnson/lesson02/notes.py
```diff
@@ -0,0 +1,96 @@
...
+
+# key word arguments or qwards
+# positional arguments
+# you can leave off the return statement
+# you can return multiple values
+#def add():
+#   return 1, 2, "blah"
+# will return it as a tuple
+# local variable = variables in scope
+# pep8???
```
HarryMaher: Not sure if you wanted me to read this, but you put it in your PR, so yeah, PEP 8 is the style python uses for the standard library and what we're trying to teach in this class. When you have a minute check this out: https://www.python.org/dev/peps/pep-0008/ 

And when you get bored of trying to read it, just install a linter that will yell at you when you're not conforming to the style.

PR #43, opened 2018-04-17
Pull request for Lesson02 mathewcm 18.04.16

mathewcmartin:


students/mathewcm/break_me.py
```diff
@@ -8,7 +8,7 @@
 #    if a > b
 
 def test4():
-   raise AttributeError
+   raise AttributeErrorg
```
PythonCHB: The idea was to write some code that would raise an AttributeError, rather than raising it yourself :-)

We’ll cover raising and handling Exceptions in a couple weeks

students/mathewcm/lesson02/fizzbuzz.py
```diff
@@ -0,0 +1,17 @@
...
+        if i % 3 == 0 and i % 5 == 0:
+            print("Fizzbuzz")
+        elif i % 3 == 0:
+            print("Fizz")
+        elif i % 5 == 0:
+            print("Buzz")
+        else:
+            print(i)
+
+fizzbuzz()
```
PythonCHB: Perfectly solid solution.

students/mathewcm/lesson02/gridPrinter.py
```diff
@@ -0,0 +1,60 @@
...
+    """
+    print(" + ----"*width,"+")
+
+def postFunc(width):
+    """
+    postFunc(p) prints 'width' number of posts wide.
+    """
+    print(" |     "*width, "|")
+
+#Need to dry code and replace length and width to just a 'units' that is length times height.
```
PythonCHB: Yup!

students/mathewcm/lesson02/gridPrinter.py
```diff
@@ -0,0 +1,60 @@
...
+    """
+    print("+----+----+", end=" ")
+
+def post():
+    """
+    post() function defines one of eight posts separating and propping up the beams in a 2x2 grid pattern
+    """
+    print("|     |     |", end=" ")
+
+def printGrid():
```
PythonCHB: It would have been nice to see some loops or string multiplication here — but you do get to that later — so good!

students/mathewcm/lesson02/series.py
```diff
@@ -0,0 +1,56 @@
...
+    a = 1
+    b = 1
+    # First two members of the series
+    series = [a, b]
+    for i in range(n):
+        c = a + b
+        series.append(c)
+        a = b
+        b = c
+    return series[-4]
```
PythonCHB: Why -4? 

You only want to compute as many as you need.

And could you do it without storing the whole list?

Also — this is an iterative vs. a recursive solution— can you do it recursively?

students/mathewcm/lesson02/series.py
```diff
@@ -0,0 +1,56 @@
...
+
+def sum_series(n, a=0, b=1):
+    """
+    This function is a generalization of the fibonacci sequence and the Lucas sequence and defaults to the values of the fibonacci sequence if no optional inputs for a and b are entered. The function returns the nth value in the series.
+    """
+    if n == 1:
+        return [1]
+    if n == 2:
+        return [1, 1]
```
PythonCHB: Shouldn’t these be using a and b?

students/mathewcm/lesson02/series.py
```diff
@@ -0,0 +1,56 @@
...
+    # First two members of the series
+    series = [a, b]
+    for i in range(n):
+        c = a + b
+        series.append(c)
+        a = b
+        b = c
+    return series[-3]
```
PythonCHB: -3 this time?

Makes me think you are trying things till it works— how would you change it so it’s the last item — series[-1]

PR #44, opened 2018-04-17
Assignment 1 Pull Request 

weepler:
Submitting a pull request for the first assignment. Please note the following:

- I was unable to get the rows to line up due to division discrepancies in step 3 of the printer gride assignment 
- I was unable to get the lucas series to pull in my sum_series function in step 3 of the fibonacci series assignment


PythonCHB:
Can we review this in class?

I don’r Want to embarrass you, but live debugging can be instructive.

students/Wayne/Assignments/Lesson2/FizzBuzzExercise.py
```diff
@@ -0,0 +1,12 @@
+#Creating FizzBuzz
+
+
+for i in range(1,101):
```
HarryMaher: This works!

students/Wayne/Assignments/Lesson2/FizzBuzzExercise.py
```diff
@@ -0,0 +1,12 @@
...
+    if i % 3 == 0 and not i % 5 == 0:
+        print('fizz')
+    elif i % 5 == 0 and not i % 3 == 0:
+        print('buzz')
+    elif i % 3 ==0 and i % 5 == 0:
```
HarryMaher: If you put this last elif higher up you could avoid the "and not" parts in your other two if/elifs.

students/Wayne/Assignments/Lesson2/FibonacciSeriesExercise.py
```diff
@@ -0,0 +1,66 @@
...
+    if lucas0 ==2 and lucas1 == 1:
+        return lucas_series(n)
+    else:
+        return fibinacci_series(n)
+
+print(sum_series1(15,2,1))
+
+# Build sum_series without lucas and fibonacci functions 
+
+def sum_series (n,lucas0=None,lucas1=None):
```
HarryMaher: Thanks for letting Chris go over this one in class!

students/Wayne/Assignments/Lesson2/PrinterGridExercise.py
```diff
@@ -0,0 +1,91 @@
+#print grid with single variable
+
+# declaring grid variables
+
+plus = '+'
+
+endplus = '+'
+
+minus = ' - '
```
HarryMaher: This'll make the squares wider, with two spaces between each dash instead of one.

students/Wayne/Assignments/Lesson2/PrinterGridExercise.py
```diff
@@ -0,0 +1,91 @@
...
+
+def print_grid(n):
+    count = n
+    centerrow = ((n)//2)+1
+    center = ((n)//2)-1
+    row = (plus + (minus*center) + plus + (minus*center) +plus)
+    column = (line + (space * center) + line + (space * center) +line)
+    
+    while count >= 1:
+        if count ==1:
```
HarryMaher: While I think the for loop approach (see an example in the solutions folder on github) is a little cleaner, I really appreciate the logic behind this. 
A couple of comments:
- -= is a little cleaner here
- if you have the same result, (i.e. "print(row)" under different elifs, it's better to combine them for a number of reasons. It's more readable and it's better for editing later if you have to change, for example, what they're printing. 
```
if count == 1 or count == n or count == center_row:
    print(row)
else:
    print(column)
count -= 1

File: students/Wayne/Assignments/Lesson2/PrinterGridExercise.py [hunk 6]
```diff
@@ -0,0 +1,91 @@
...
+    mcolumn = (line + (space *m))
+    endcolumn = (mcolumn + endline)
+    newcolumn = (mcolumn + endline)
+    
+    while unitcount >= 1:
+        if unitcount ==1:
+            print(newrow + ((newrow[1:])*(n-1)))
+        elif unitcount == m:
+            print(newrow + ((newrow[1:])*(n-1)))
+        elif unitcount % m == 0:
```
HarryMaher: You had it 99% there. The "unitcount % m" is why it's going to print a newrow after 2 columns instead of 3 for a size 3 (when we want 3).
This is how it'd look with that small bit fixed
```
def print_grid2(n,m):
    unitcount = (m * n)+(n)  # Also took out +1 here
    mrow = (plus + (minus*m))
    newrow = mrow + endplus
    endrow = endplus
    mcolumn = (line + (space *m))
    endcolumn = (mcolumn + endline)
    newcolumn = (mcolumn + endline)
    while unitcount >= 0:  # Now we want the 0th instance to account for below
        if unitcount % (m+1) == 0:  # put +1, set it to 0
            print(newrow + ((newrow[1:])*(n-1)))
        else:
            print(newcolumn + ((newcolumn[1:])*(n-1)))
        unitcount -= 1
```

PR #45, opened 2018-04-19
Lesson 3 Assignments - Wes Wang

WesWang93:
Slicing lab incomplete & string format missing task 6

### DIFF
--- a/students/WesWang/lesson3/mailroom_part1.py
+++ b/students/WesWang/lesson3/mailroom_part1.py
@@ -1,0 +1,86 @@
+#=========================================================================================
+#Name: Mailroom
+#Author: Wesley Wang
+#Date: 4/18/2018
+#=========================================================================================
+
+"""
+Mailroom Part1
+"""
+import sys
+
+donors = [["Donor A", 140000, 2],
+          ["Donor B", 1000, 1],
+          ["Donor C", 3000, 3],
+          ["Donor D", 20000, 4]]
+
+def thank_letter(donor, amount):
+  print("-"*80)
+  print(f"Dear {donor},\n\n\tWe have received you donation of ${amount:,.2f}. We greatly appreciate your generous contribution. This donation is meaningful and crucial to our organization's misson.\n\n\nSincerely,\n\nWesley Wang\n")
+  print("-"*80)
+
+def thank_you():
+    # Ask for user input
+    response = input("Please enter donor's full name, or type 'list' to check all donors: ").title()
+    if response.lower() == "list":
+      print("-"*20 + "\nList of Donors:\n" + "-"*20)
+      for donor in donors:
+        print(donor[0])
+      print("-"*20)
+      thank_you()
+    amount = int(input("Please enter donation amount for " + response + ":"))
+    for donor in donors:
+      if response.lower() == donor[0].lower():
+        donor[1] += amount
+        donor[2] += 1
+        thank_letter(donor[0], amount)
+        print("Return to main menu!!")
+        mainloop()
+    donors.append([response, amount, 1])
+    thank_letter(response, amount)
+    mainloop()
+
+
+def report():
+  #header = "Donor Name" + " "*30 + "| Total Given | Num Gifts | Average Gifts |"
+  header = "{:30}| {:15}| {:10}| {:20}".format("Donor Name","Total Given", "Num Gifts","Average Gifts")
+  print("-"*75)
+  print(header)
+  print("-"*75)
+
+  for donor in donors:
+    if len(donor[0]) > 30:
+      print("Error: Donor name exceeds length limit! Please verify " + donor[0] + "'s information")
+    elif len(str(donor[1])) > 9:
+      print("Error: Donated amount exceeds limit (6 digits)! Please verify " + donor[0] + "'s information!")
+    elif len(str(donor[2])) > 10:
+      print("Error: Donation number exceeds limit! Please verify " + donor[0] + "'s information!")
+    else:
+      print(f"{donor[0]:30}${donor[1]:>15,.2f}{donor[2]:>13}{donor[1]/donor[2]:>15,.2f}")
+  print("-"*75)
+
+def quit():
+    print("You have successfully quit Mailroom")
+    sys.exit()
+
+def mainloop():
+    print("Welcome to the Mailroom")
+
+    response = ""
+    while response != "3":
+        print("What do you want to do?")
+        response = input("[1] Thank you\n[2] Report\n[3] Quit\n")
+        print(response, " was chosen")
+
+        if response not in ("1","2","3"):
+            print("Not a Valid Response --- Type 1,2,3")
+        elif response == "1":
+            thank_you()
+        elif response == "2":
+            report()
+        else:
+            quit()
+
+
+if __name__ == "__main__":
+    mainloop()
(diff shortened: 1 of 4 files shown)


students/WesWang/lesson3/mailroom_part1.py
```diff
@@ -0,0 +1,86 @@
...
+  print("-"*75)
+  print(header)
+  print("-"*75)
+
+  for donor in donors:
+    if len(donor[0]) > 30:
+      print("Error: Donor name exceeds length limit! Please verify " + donor[0] + "'s information")
+    elif len(str(donor[1])) > 9:
+      print("Error: Donated amount exceeds limit (6 digits)! Please verify " + donor[0] + "'s information!")
+    elif len(str(donor[2])) > 10:
```
PythonCHB: why limit the number of donors?

students/WesWang/lesson3/mailroom_part1.py
```diff
@@ -0,0 +1,86 @@
...
+def report():
+  #header = "Donor Name" + " "*30 + "| Total Given | Num Gifts | Average Gifts |"
+  header = "{:30}| {:15}| {:10}| {:20}".format("Donor Name","Total Given", "Num Gifts","Average Gifts")
+  print("-"*75)
+  print(header)
+  print("-"*75)
+
+  for donor in donors:
+    if len(donor[0]) > 30:
+      print("Error: Donor name exceeds length limit! Please verify " + donor[0] + "'s information")
```
PythonCHB: you are referencing each index more than once -- probably better to pull it out and give it a name:

name = donor[0]

then use name in the test, and formatting.

PythonCHB: why not use an f-string here, too?

students/WesWang/lesson3/mailroom_part1.py
```diff
@@ -0,0 +1,86 @@
...
+#Author: Wesley Wang
+#Date: 4/18/2018
+#=========================================================================================
+
+"""
+Mailroom Part1
+"""
+import sys
+
+donors = [["Donor A", 140000, 2],
```
PythonCHB: you want to store all the donations, not just the total and number. -- you can calculate the total and number of donations when you need them.

students/WesWang/lesson3/mailroom_part1.py
```diff
@@ -0,0 +1,86 @@
...
+
+def quit():
+    print("You have successfully quit Mailroom")
+    sys.exit()
+
+def mainloop():
+    print("Welcome to the Mailroom")
+
+    response = ""
+    while response != "3":
```
PythonCHB: I think you'll never get to response == "3" --it will quite out instead. So a plain "while True" would work here.
And then you don't need to initialize response.

students/WesWang/lesson3/mailroom_part1.py
```diff
@@ -0,0 +1,86 @@
...
+
+def thank_letter(donor, amount):
+  print("-"*80)
+  print(f"Dear {donor},\n\n\tWe have received you donation of ${amount:,.2f}. We greatly appreciate your generous contribution. This donation is meaningful and crucial to our organization's misson.\n\n\nSincerely,\n\nWesley Wang\n")
+  print("-"*80)
+
+def thank_you():
+    # Ask for user input
+    response = input("Please enter donor's full name, or type 'list' to check all donors: ").title()
+    if response.lower() == "list":
```
PythonCHB: probably a good idea to make the listing a separate function.

students/WesWang/lesson3/mailroom_part1.py
```diff
@@ -0,0 +1,86 @@
...
+        elif response == "1":
+            thank_you()
+        elif response == "2":
+            report()
+        else:
+            quit()
+
+
+if __name__ == "__main__":
+    mainloop()
```
PythonCHB: nice job overall with mailroom  -- all the logic is there.

students/WesWang/lesson3/slicing_lab.py
```diff
@@ -0,0 +1,38 @@
+#=========================================================================================
+#Name: Slicing Lab
+#Author: Wesley Wang
+#Date: 4/18/2018
+#=========================================================================================
+
+# Exchange first and last items
+def exchange_first_last(seq):
+  result = [seq[-1]] + seq[1:-1] + [seq[0]]
```
PythonCHB: better to use a length-one slice:
seq[-1:]

that will give you back a sequence of the same type.

This code will give you a list, which, if you are slicing, say, a string, would not be what you want!

students/WesWang/lesson3/slicing_lab.py
```diff
@@ -0,0 +1,38 @@
...
+
+# Reverse order of items with slicing
+def reverse(seq):
+  result = seq[::-1]
+  return result
+
+#
+def mid_last_first(seq):
+    first_third = len(seq)//3
+    second_third = (len(seq)//3)*2
```
PythonCHB: `first_third * 2` why take the length twice?

students/WesWang/lesson3/slicing_lab.py
```diff
@@ -0,0 +1,38 @@
...
+    result = seq[first_third:second_third] + seq[second_third:] + seq[:first_third]
+    return result
+
+a_string = "This is a string"
+a_tuple = [1,2,3,4,5,6,7,8,9,10,11,12]
+
+assert exchange_first_last(a_string) == "ghis is a strinT"
+assert exchange_first_last(a_tuple) == [12,2,3,4,5,6,7,8,9,10,11,1]
```
PythonCHB: this all looks great.

students/WesWang/lesson3/string_format.py
```diff
@@ -0,0 +1,40 @@
...
+
+#String Formatting Lab
+
+tuple1 = ( 2, 123.4567, 10000, 12345.67)
+tuple3 = (1,2,3,4,5,6)
+tuple4 = ( 4, 30, 2017, 2, 27)
+list5 = ['oranges', 1.3, 'lemons', 1.1]
+
+def task1(tup):
+  result = "file_{0}: {1:.2f}, {2:.2e}, {3:.2e}".format(*tup)
```
PythonCHB: nice! you discovered `*args`!

students/WesWang/lesson3/string_format.py
```diff
@@ -0,0 +1,40 @@
...
+  return result
+
+def task2(tup):
+  result = "file_{0:003}: {1:.0f}, {2:.1e}, {3:.0f}".format(*tup)
+  return result
+
+def task3(tup):
+  form_string = ", {:d}"*(len(tup)-1)
+  result = f"The {len(tup)} numbers are {tup[0]}" + form_string.format(*tup[1:])
```
PythonCHB: why put tup[0] here?

students/WesWang/lesson3/string_format.py
```diff
@@ -0,0 +1,40 @@
...
+def task5(lst):
+  result = f"The weight of an {lst[0][:-1].upper()} is {lst[1]} and the weight of a {lst[2][:-1].upper()} is {lst[3]}"
+  return result
+
+
+print("Task 1: " + task1(tuple1))
+print("Task 2: " + task2(tuple1))
+print("Task 3: " + task3(tuple3))
+print("Task 4: " + task4(tuple4))
+print("Task 5: " + task5(list5))
```
PythonCHB: nicely done!

PR #46, opened 2018-04-19
lesson03 

tammytdo:


students/tammyd/lesson02/series.py
```diff
@@ -43,4 +47,46 @@ def sum_series(required, *args, **kwargs):
...
-sum_series()
+"""
+
+def sum_series(n, s0=None, s1=None):
+    print("sum_series called:", n)
+    if n == 0:
+        return s0
+    elif n == 1:
+        return s1
+    elif n == 2:
```
PythonCHB: This was an experiment that failed in class :-(

so be careful -- you only need the first two numbers specified. and they ARE required.

students/tammyd/lesson02/series.py
```diff
@@ -31,6 +33,8 @@ def lucas(n):
 #I am having a hard time understanding this part. I will ask during class. 
 
 """This function called sum_series has one required parameter and two optional parameters. The required parameter determines which element in the series to print. The two optional parameters have default values of 0 and 1 and determine the first two values for the series to be produced."""
+"""
+
 def sum_series(required, *args, **kwargs):
 
     if x == 0 and y == 1:
```
PythonCHB: ahh -- this is backwards from what we want -- you want to write a generic one, where you specify x and y, and then you can write a specific one, where x and y are fixed.

See the Solutions.

PR #47, opened 2018-04-19
Lesson 3: Pushing new changes

meet29in:


students/meet_shah/lesson03/slicing.py
```diff
@@ -0,0 +1,15 @@
+a_string = "this is a string"
+a_tuple = (2, 54, 13, 12, 5, 32)
+
+def exchange_first_last(seq):
+   first = seq[0]
+   middle = seq[1:len(seq)-1]
```
PythonCHB: You never need len(). try:

seq[1:-1]

PythonCHB:
No real need for the intermediate variable names — but you got the slicing right — which is the point.

PR #48, opened 2018-04-20
Andy Kwon - Lesson03 and 04 PR

KwonAndy:
This is what I have thus far. Still working on it.


PythonCHB:
you've got conflicts in your series.py file.

Try merging it all together and fixing teh conflicts on your machine:

```
$ git pull

$ git pull upstream master
```

you should get a message from git that there is a conflict -- edit the file and fix it, then comit everything again.

$ git commit -a
$ git push

and the conflict should go away in the PR.

students/andykwon/lesson02/series.py
```diff
@@ -0,0 +1,116 @@
...
+
+print(sum_series(7, 2, 4))
+print(lucas(7))
+print(fibonacci(7))
+
+#Tests:
+
+import unittest
+
+class TestSeriesNumbers(unittest.TestCase):
```
PythonCHB: awesome unit test!

we'll be getting to that soon, but great to thinking that way!

students/andykwon/lesson03/list_lab.py
```diff
@@ -0,0 +1,37 @@
...
+
+print(fruit)
+
+print("fruits that only start with P:")
+
+# Printing fruits in the list that start with "P"
+for i in range (len(fruit)):
+    if fruit[i][0] == "P":
+
+        print(fruit[i])
```
PythonCHB: this all looks good.

students/andykwon/lesson03/mailroom.py
```diff
@@ -0,0 +1,103 @@
...
+
+    print("== Send a Thank You ==\n")
+
+    print("Please enter the name of the person you would like to send a Thank You to."
+          "(type 'list' if you would like a list of the donor names\n")
+
+    response_name = input(">>")
+
+    if response_name == "list":
+        print(donors)
```
PythonCHB: you'll want to write a function that prints only the donor names.

students/andykwon/lesson03/mailroom.py
```diff
@@ -0,0 +1,103 @@
...
+
+        donors.append((response_name, []))
+
+    # Ask for donation amount; add donation amount to associated donation list
+    print("Please enter the donation amount:\n")
+
+    response_donation = input(">>")
+
+    for item in donors:
+        if response_name in item:
```
PythonCHB: you've now looped through the list twice looking for the donor. Can you avoid that?

students/andykwon/lesson03/mailroom.py
```diff
@@ -0,0 +1,103 @@
...
+    print("Are you sure you want to quit?\n"
+          "1) Yes, quit"
+          "2) No, don't quit")
+
+    response = input(">")
+
+    if response == "1":
+        sys.exit()
+    elif response == "2":
+        mainloop()
```
PythonCHB: careful here -- this will call the mainloop recursivly:

mainloop() calls quit(), quit() calls mainloop(), mainloop calls quit() .... etc, potentially a LOT of times.

so better for quit() to return -- then make sure that mainloop() will go back through the loop when it does.

students/andykwon/lesson03/mailroom.py
```diff
@@ -0,0 +1,103 @@
...
+
+
+def mainloop():
+
+    response = ""
```
PythonCHB: don't need to initialize response here.

students/andykwon/lesson03/mailroom.py
```diff
@@ -0,0 +1,103 @@
...
+            print("Not a valid response")
+        elif response == "1":
+            thank_you()
+            continue
+        elif response == "2":
+            report()
+            continue
+        elif response == "3":
+            quit()
+            continue
```
PythonCHB: this will continue the loop if quit returns -- so all you need to do is get rid of the call to mainloop in quit()

students/andykwon/lesson03/slicing_lab.py
```diff
@@ -0,0 +1,75 @@
...
+    print("this is every_other_removed")
+
+    return seq[::2]
+
+def remove_first_last_four_and_every_other(seq):
+    print("this is remove_first_last_four_and_every_other")
+
+    seq = seq[4:]
+    seq = seq[:-4]
+    seq = seq[::2]
```
PythonCHB: you can do this with all in one slice....

students/andykwon/lesson03/slicing_lab.py
```diff
@@ -0,0 +1,75 @@
...
+
+def elements_reversed_slicing(seq):
+    print("this is elements_reversed_slicing")
+
+    return seq[::-1]
+
+def reorder_middle_last_first_third(seq):
+    print("this is reorder_middle_last_first_third")
+
+    first_third = seq[:int(len(seq)/3)]
```
PythonCHB: `len(seq) // 3`

will return an int saving that call.

and good idea to store it, so you don't need to call len() a whole bunch of times.

PythonCHB:
I"m not going to merge 'till the conflict is resolved.....

KwonAndy:
Hello Chris,


I believe the conflict has been resolved.


Thank you,


Andy Kwon


________________________________

PR #49, opened 2018-04-20
Lesson 03

DwightJay:
Adding what I have so far.


students/jayjohnson/lesson02/Grid_Printer.py
```diff
@@ -0,0 +1,151 @@
...
+|               |               |
+|               |               |
+|               |               |
+|               |               |
++ - - - - - - - + - - - - - - - +
+This problem is underspecified. Do something reasonable.
+'''
+
+def print_grid(n):
+    grid_count_upper = math.floor(n/2)
```
PythonCHB: n // 2 will give the same result— I. Fact, it’s called “floor division”

students/jayjohnson/lesson01/break_me.py
```diff
@@ -0,0 +1,37 @@
...
+
+# error functions
+def nameError_fuction():
+    a = 5
+    del a
+    a
+
+def typeError_fuction():
+    type = 'abc'
+    type('efg')
```
PythonCHB: So this is redefining the name “type” — be careful about redefining buildings!

Any name would have the same effect — you can’t cal a string.

students/jayjohnson/lesson01/break_me.py
```diff
@@ -0,0 +1,37 @@
...
+
+def syntaxError_fuction():
+    #del = 'error please'
+
+def AttributeError_fuction():
```
PythonCHB: This isn’r Doing anything.. ??

students/jayjohnson/lesson01/break_me.py
```diff
@@ -0,0 +1,37 @@
...
+    pass
+
+# will need to comment out all others to get thewe to fail correctly
+nameError_fuction()
+
+typeError_fuction()
+
+syntaxError_fuction()
+
+AttributeError_fuction(error)
```
PythonCHB: NameError here?

Or an error for passing an argument to a function that doesn’t take one.

Can you make it give you an AttributeError?

students/jayjohnson/lesson02/Fizz_Buzz.py
```diff
@@ -0,0 +1,28 @@
...
+x For numbers which are multiples of both three and five print “FizzBuzz” instead.
+'''
+# loop logic
+guess = False
+
+# count set to 100
+test_num = 100
+
+# loop
+while guess == False:
```
PythonCHB: Where is guess used?

DwightJay:
Lesson03 almost complete

PythonCHB:
I'm going to resolve this conflict -- not sure why it's still there, I thought we fixed it!

PR #51, opened 2018-04-20
finish assignment for lesson03

rachelz11:


students/Ruohan/lesson03/mailroom.py
```diff
@@ -0,0 +1,95 @@
...
+''' simple mailroom proram for a non profit'''
+
+import sys
+
+datas = [['William Gates III',  653784.39, 2],
+         ['Mark Zuckerberg', 16396.10, 3],
+         ['Jeff Bezos', 877.33, 1],
+         ['Paul Allen', 708.42, 3]]
+
+'''make a donors' name list'''
```
hosungs: Use `#` for a non-docstring comment. This is not a docstring comment, because it's neither for the module, nor for a function/class. Just `# make a donor's name list` should be fine.

students/Ruohan/lesson03/mailroom.py
```diff
@@ -0,0 +1,95 @@
...
+donors = []
+for data in datas:
+    donors.append(data[0])
+
+'''find if donor is already in the datas, if not add the name to the datas'''
```
hosungs: If you want this to be a docstring comment for this function, move this line below line 18, indented.

students/Ruohan/lesson03/mailroom.py
```diff
@@ -0,0 +1,95 @@
...
+def find_donor(donor):
+    donor = donor.rstrip().title()
+    if donor in donors:
+        return donor
+    else:
+        datas.append([donor, 0, 0])
```
hosungs: What would happen `find_donor("Steve Ballmer")` is called once, his donation amount was increased, and `find_donor("Steve Ballmer")` is called again?

students/Ruohan/lesson03/slicing_lab.py
```diff
@@ -0,0 +1,53 @@
...
+
+assert exchange_first_last(a_string) == 'nhis is Ruohat'
+assert exchange_first_last(a_tuple) == (19, 34, 44, 27, 56, 4, 12)
+
+
+
+def every_other_removed(seq):
+    '''return sequence with other item removed'''
+
+    if len(seq) > 1:
```
hosungs: Do you need this? What if an empty sequence is passed?

students/Ruohan/lesson03/slicing_lab.py
```diff
@@ -0,0 +1,53 @@
...
+        a_new_sequence = seq[::2]
+    return a_new_sequence
+assert every_other_removed(a_string) == 'ti sRoa'
+assert every_other_removed(a_tuple) == (12, 44, 56, 19)
+
+
+def first_last4_removed(seq):
+    '''return sequence with the first and last 4 items removed,
+    and every other item in between'''
+    if len(seq) > 4:
```
hosungs: Do you need this? What if you don't have this if condition but just return `seq[1:-4]` all the time?

students/Ruohan/lesson03/mailroom.py
rachelz11: the 'find_donor' function only used for find if the donor is already in the datas. If he is not, add him to the datas. Then use 'upgrade_dates' fuction to change the donation amount.

hosungs: But aren't you searching the `donors` list, not the `datas` list, and the `donors` list won't get any newly added donor?

rachelz11: Oh.  I see. I will fix this problem!  Thanks Dr. Song!

students/Ruohan/lesson02/fizzbuzz.py
```diff
@@ -1,9 +1,21 @@
 for i in range(1, 101):
-    if i%3==0 and i%5==0:
+    if i % 3==0 and i % 5==0:
         print('FizzBuzz')
-    elif i%3==0:
+    elif i % 3==0:
         print('Fizz')
-    elif i%5==0:
+    elif i % 5==0:
```
PythonCHB: nice PEP 8 cleanup!

students/Ruohan/lesson03/mailroom.py
```diff
@@ -0,0 +1,96 @@
+#! /usr/bin/env python3
+
+''' simple mailroom proram for a non profit'''
+
+import sys
+
+datas = [['William Gates III',  653784.39, 2],
```
PythonCHB: you're storing the total and the number of donations -- while that's all we are asking for in the report, you may well want to keep the donation history -- better to keep a list of all donations, and compute the total when you need it.

students/Ruohan/lesson03/strformat_lab.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+
+#task theree
+print('=============task three===============')
+def formatter(in_tuple):
+    q = len(in_tuple)
+    numbers = []
+    for i in range(q):
+        numbers.append('{:d}')
```
PythonCHB: this works fine -- bout could you have done it with multiplication?

students/Ruohan/lesson03/strformat_lab.py
```diff
@@ -0,0 +1,52 @@
...
+def task_four(tuple):
+    new_str = '{3:0>2d},{4},{2},{0:0>2d},{1}'.format(*tuple)
+    return new_str
+another_tuple = (4, 30, 2017, 2, 27)
+print(task_four(another_tuple))
+#task five
+print('==============task five=================')
+
+def task_five(e_list,w = 1):
+    li_new =f'The wight of an {e_list[0][:-1].upper()} is {e_list[1]*w} and the weight of a {e_list[2][:-1].upper()} is {e_list[-1]*w}'
```
PythonCHB: nice!

f-strings are pretty nifty, eh?

students/Ruohan/lesson04/kata.py
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env python3
+
+''' trigrams '''
+import string
+import random
+
+file_name = 'sherlock.txt'
+with open (file_name) as f_handle:
```
PythonCHB: you don't want to have a space between a function and the parens...

students/Ruohan/lesson04/kata.py
```diff
@@ -0,0 +1,43 @@
...
+# find the real content we need from the file
+    line_useful = ''
+    bool_flag = False
+    while True:
```
PythonCHB: could you do this as a for loop?

```
for line in f_handle:
   bool_flag = ...
```

students/Ruohan/lesson04/kata.py
```diff
@@ -0,0 +1,43 @@
...
+
+''' trigrams '''
+import string
+import random
+
+file_name = 'sherlock.txt'
+with open (file_name) as f_handle:
+# find the real content we need from the file
+    line_useful = ''
+    bool_flag = False
```
PythonCHB: try to use variable names that mean something, like "valid_line" or "good_data" or whatever, the fact that its a boolean flag doesn't tell us anything about what its meaning is.

students/Ruohan/lesson04/kata.py
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env python3
+
+''' trigrams '''
+import string
+import random
+
+file_name = 'sherlock.txt'
+with open (file_name) as f_handle:
+# find the real content we need from the file
+    line_useful = ''
```
PythonCHB: line_useful will be all the lines eventually, yes? so maybe something like "text", "all_text" "all_lines" "good_lines", or even "useful_lines"....

students/Ruohan/lesson04/kata.py
```diff
@@ -0,0 +1,43 @@
...
+    bool_flag = False
+    while True:
+        line = f_handle.readline()
+        bool_flag = bool_flag or (line == 'To Sherlock Holmes she is always THE woman. I have seldom heard\n')
+        if bool_flag:
+            line_useful += line
```
PythonCHB: building up a big string by adding parts on bit by bit is inefficient in Python. 

Better to build it up in a list, and then `.join()` the list.

students/Ruohan/lesson04/kata.py
```diff
@@ -0,0 +1,43 @@
...
+        if line == 'Walsall, where I believe that she has met with considerable success.\n':
+            break
+
+# deltet the punctuation
+
+table = str.maketrans(dict.fromkeys(list(string.punctuation)))
+line_useful = line_useful.strip().translate(table)
```
PythonCHB: nice job finding `.translate`!

students/Ruohan/lesson04/kata.py
```diff
@@ -0,0 +1,43 @@
...
+    datas.setdefault(key,[]).append(words[i+2])
+
+def new_text():
+    k1 = input('Enter the first word>>')
+    k2 = input('Enter the second word>>')
+    k = (k1,k2)
+    new = ''
+    while True:
+        if k in datas:
+            new_word = str(random.choice(datas.get(k)))
```
PythonCHB: if you know it's there, you can use indexing:

`new_word = random.choice(datas[k])`

(and it should be  a string already)

but you could also use get and not test for "in" first.

followers = datas.get(k)
if followers:
    new_word = random.choice(followers)
    ...
else:
    break

students/Ruohan/lesson04/kata.py
```diff
@@ -0,0 +1,43 @@
...
+for i in range(len(words)-2):
+    key = (words[i], words[i+1])
+    datas.setdefault(key,[]).append(words[i+2])
+
+def new_text():
+    k1 = input('Enter the first word>>')
+    k2 = input('Enter the second word>>')
+    k = (k1,k2)
+    new = ''
+    while True:
```
PythonCHB: might this go on forever? probably good to have a stopping point....

students/Ruohan/lesson04/kata.py
```diff
@@ -0,0 +1,43 @@
...
+line_useful = line_useful.strip().translate(table)
+
+words = line_useful.split()
+datas = {}
+for i in range(len(words)-2):
+    key = (words[i], words[i+1])
+    datas.setdefault(key,[]).append(words[i+2])
+
+def new_text():
+    k1 = input('Enter the first word>>')
```
PythonCHB: you should grab the first pair of words from the dict somehow -- your user might not choose two that exist in the text....

students/Ruohan/lesson04/mailroom_part2
```diff
@@ -0,0 +1,3 @@
+#! /usr/bin/env python3
```
PythonCHB: this looks like an error -- you should remove it from git:

`git rm mailroom_part2`

students/Ruohan/lesson04/mailroom_part2.py
```diff
@@ -0,0 +1,115 @@
...
+         ['Mark Zuckerberg', 16396.10, 3],
+         ['Jeff Bezos', 877.33, 1],
+         ['Paul Allen', 708.42, 3]]
+
+# make a new data structure for part2
+datas = []
+
+for i in old_datas:
+    dic = {}
+    dic.setdefault('name', i[0])
```
PythonCHB: do you need setdefault here? why not just set it:

dic['name] = i[0]

you could use some nifty tricks here:

`dic = dict(zip(('name', 'donation_amount', 'donation_times'), i))`

(untested)

students/Ruohan/lesson04/mailroom_part2.py
```diff
@@ -0,0 +1,115 @@
+#! /usr/bin/env python3
+
+import sys
+import string
+
+old_datas = [['William Gates III',  653784.39, 2],
```
PythonCHB: better to store all old donations, and compute sum when you need it. So if you wanted other info, you could get it.

students/Ruohan/lesson04/mailroom_part2.py
```diff
@@ -0,0 +1,115 @@
...
+    dic = {}
+    dic.setdefault('name', i[0])
+    dic.setdefault('donation_amount', i[1])
+    dic.setdefault('donation_times',i[2])
+    dic.setdefault('recent_donation')
+    datas.append(dic)
+
+
+#make a donors' name list
+def donors_name_list():
```
PythonCHB: nice opportunity for a list comprehension here.

students/Ruohan/lesson04/mailroom_part2.py
```diff
@@ -0,0 +1,115 @@
...
+    donors = []
+    for data in datas:
+        donors.append(data['name'])
+    return donors
+
+
+def find_donor(donor,donors):
+    '''find if donor is already in the datas, if not add the name to the datas'''
+    if donor not in donors:
+        donor_info = {'name': donor}
```
PythonCHB: you can do this with one literal:

```
donor_info = {'name': donor,
              'donation_amount':0,
              'donation_times': 0}
```

students/Ruohan/lesson04/mailroom_part2.py
```diff
@@ -0,0 +1,115 @@
...
+
+
+def upgrade_datas(donor):
+    # upgrade donors' total donation amount, donation times and recent donation amount
+    amount = float(input('Enter donation amount: '))
+    for data in datas:
+        if donor == data.get('name'):
+            data['donation_amount'] +=  amount
+            data['donation_times'] += 1
+            data.setdefault('recent_donation', amount)
```
PythonCHB: use regular dict indexing here:

`data['recent_donation', amount]`

`.setdefault` is for when you want to use the result, and it might already be in there.

PR #54, opened 2018-04-23
OSiddiquee Mailroom

osiddiquee:
PR for mailroom.py


students/Osiddiquee/lesson03/mailroom_report.py
```diff
@@ -0,0 +1,21 @@
...
+
+def report():
+    report = []
+    for (donor, amounts) in donors:
+        report.append((donor, sum(amounts), len(amounts), sum(amounts)/len(amounts)))
+    report.sort(key = totaldonations, reverse = True)
+    print('Donor Name',' '*10, '|  Total Given  |', ' Num Gifts ', '|  Average Gift  |')
+    print('-'*70)
+    for (donor, total, num, avg) in report:
+        print(donor, ' '*(21 - len(str(donor))), #print donor
```
hosungs: Use string `.format()`. This line can be `"{:10}".format(donor)`. See https://pyformat.info/ and update this line and the following 3 lines as well.

PR #57, opened 2018-04-23
Lesson4 Submission

WesWang93:
Trigram + Mail Room Part2

### DIFF
--- a/students/WesWang/lesson4/mailroom_part2.py
+++ b/students/WesWang/lesson4/mailroom_part2.py
@@ -1,0 +1,83 @@
+#=========================================================================================
+#Name: Mailroom
+#Author: Wesley Wang
+#Date: 4/23/2018
+#=========================================================================================
+
+"""
+Mailroom Part2
+"""
+import sys
+
+donors = {"Donor A":[100000, 40000]
+        , "Donor B":[1000]
+        , "Donor C":[1000,1000,1000]
+        , "Donor D":[5000,10000,5000]}
+
+
+def thank_letter(donor, amount):
+    print("-"*80)
+    print(f"Dear {donor},\n\n\tWe have received you donation of ${amount:,.2f}. We greatly appreciate your generous contribution. This donation is meaningful and crucial to our organization's misson.\n\n\nSincerely,\n\nWesley Wang\n")
+    print("-"*80)
+
+def thank_you():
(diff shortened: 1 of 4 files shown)


students/WesWang/lesson4/mailroom_part2.py
```diff
@@ -0,0 +1,83 @@
...
+def thank_you():
+    # Ask for user input
+    response = input("Please enter donor's full name, or type 'list' to check all donors: ").title()
+    if response.lower() == "list":
+        print("-"*20 + "\nList of Donors:\n" + "-"*20)
+        for donor in donors.keys():
+            print(donor)
+        print("-"*20)
+        thank_you()
+    amount = int(input("Please enter donation amount for " + response + ":"))
```
HarryMaher: Casing the amount to an int() makes it so that if someone donates "57.34" throw an error. So we only allow full dollar amounts, which is fine if that's what you want, just be aware that you can't accept non-whole dollar donations

students/WesWang/lesson4/mailroom_part2.py
```diff
@@ -0,0 +1,83 @@
...
+    mainloop()
+
+
+def report():
+    header = "{:30}| {:15}| {:10}| {:20}".format("Donor Name","Total Given", "Num Gifts","Average Gifts")
+    print("-"*75)
+    print(header)
+    print("-"*75)
+
+    for donor in donors.keys():
```
HarryMaher: Just a note, your code using donors.keys()
returns the same as:
```
for donor in donors:
```

students/WesWang/lesson4/mailroom_part2.py
```diff
@@ -0,0 +1,83 @@
...
+    response = ""
+    while True:
+        print("What do you want to do?")
+        response = input("[1] Thank you\n[2] Report\n[3] Quit\n")
+        print(response, " was chosen")
+
+        if response not in ("1","2","3"):
+            print("Not a Valid Response --- Type 1,2,3")
+        elif response == "1":
+            thank_you()
```
HarryMaher: Check out using dictionaries for this so you don't have to keep adding elifs. Dicts are more efficient and look cleaner. And you can store the name of a function (without quotes) as the value of the dictionary to be able to call functions using a dictionary

PR #60, opened 2018-04-26
osiddiquee lesson04/lesson05

osiddiquee:
PR for lesson04 and lesson05

### DIFF
--- a/students/Osiddiquee/lesson04/trigram.py
+++ b/students/Osiddiquee/lesson04/trigram.py
@@ -1,0 +1,42 @@
+import random
+
+#reading in a text file
+inputfile = open('sherlock_small.txt')
+novel = inputfile.read()
+inputfile.close()
+
+# Builds the dicitonary with which the Kata is built
+def dictionary_builder(text):
+    words = text.replace('\n', ' ').split(' ')
+    trigrams = {}
+    for i, word in enumerate(words):
+        if i == (len(words) - 2):
+            break
+        elif words[i] + ' ' + words[i + 1] not in trigrams:
+            trigrams[words[i] + ' ' + words[i + 1]] = [words[i + 2]]
+        else:
+            trigrams[words[i] + ' ' + words[i + 1]].append(words[i + 2])
+    return trigrams
+
+
+def random_trigram(list):
+    return random.choice(list)
+
+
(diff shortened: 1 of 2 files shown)


students/Osiddiquee/lesson04/trigram.py
```diff
@@ -0,0 +1,34 @@
+import random
+
+#reading in a text file
+inputfile = open('sherlock_small.txt')
+novel = inputfile.read()
+inputfile.close()
+
+# Builds the dicitonary with which the Kata is built
+def dictionary_builder(text):
+    words = text.replace('\n', ' ').split(' ')
```
HarryMaher: .split() with no arguments splits all whitespace including "\n" chars. Note that by replacing with " " means that you'll be putting a lot of whitespace in. Then since you're splitting by " " (one space) if  you have something like:
```
In [39]: print("these are words \n\n with \n spaces".replace("\n"," ").split(" "))
['these', 'are', 'words', '', '', '', 'with', '', '', 'spaces']

In [40]: print("these are words \n\n with \n spaces".split())
['these', 'are', 'words', 'with', 'spaces']
```

So it's better to just .split() instead of replacing and splitting by a single whitespace so you don't get so many empty strings in your dict

students/Osiddiquee/lesson04/trigram.py
```diff
@@ -0,0 +1,34 @@
+import random
+
+#reading in a text file
+inputfile = open('sherlock_small.txt')
```
HarryMaher: We wanna use a "with" context manager statement to open files:
```
with open('sherlock_small.txt') as file:
    novel = file.read()
```
This closes the file once you're de-indented so you don't need file.close().

students/Osiddiquee/lesson04/trigram.py
```diff
@@ -0,0 +1,34 @@
...
+        else:
+            trigrams[words[i] + ' ' + words[i + 1]].append(words[i + 2])
+    return trigrams
+
+
+def random_trigram(list):
+    return random.choice(list)
+
+
+def text_generator(text, start, num_words = 10):
```
HarryMaher: I like the "start" idea! Cool

PR #63, opened 2018-04-27
Completed lesson04 exercises

ssoukasene:
finished mailroom_2.py and trigrams.py

### DIFF
--- a/students/stephen/lesson05/except_exercise.py
+++ b/students/stephen/lesson05/except_exercise.py
@@ -17,19 +17,15 @@ first_try = ['spam', 'cheese', 'mr death']
 try:
     joke = fun(first_try[0])
 except NameError:
-    print("That joke doesn't exists")
+    joke = fun(first_try[1])
 
 # Here is a try/except block. Add an else that prints not_joke
 try:
     not_joke = fun(first_try[2])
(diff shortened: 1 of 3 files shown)


ssoukasene:
Did exceptions exercise.
Also, did mailroom_3.py using exceptions, but not sure where to apply list comprehensions.

students/stephen/lesson05/except_exercise.py
```diff
@@ -49,4 +45,11 @@
...
 langs = ['java', 'c', 'python']
 
-more_joke = more_fun(langs[0])
+try:
+    more_joke = more_fun(langs[0])
+except IndexError:
+    more_joke = more_fun(langs[1])
+else:
+    more_fun(langs[2])
+finally:
```
HarryMaher: Nice, looks good!

PythonCHB:
@ssoukasene: There may not be any good places for a comprehension in your code -- so don't worry about it.

However, I see this:

```
def names():
    """
    Refreshes the names list with current donors
    """
    names = []
    for tup in donors:
        names.append(tup[0])
    return names
```
which is begging for a comprehension! (though I don't think it does what you expect with a dict as the donors data structure)

PR #64, opened 2018-04-28
MailRoomPt1 Update

ahwall78:


students/ahwall/Lesson05Homework/MailRoomPt3.py
```diff
@@ -0,0 +1,151 @@
...
+        response = input('>>')
+
+# could add a exception here but the if reponse covers it nicely, but it more
+# options are added it may be necessary to include an exception
+        if response not in ("1", "2", "3", "4"):
+            print("***Not a Valid response  -- type 1,2,3***\n")
+            continue
+        elif response == "1":
+            thank_you()
+        elif response == "2":
```
HarryMaher: This should be a dict, I think I've said that above. Check out the "solutions" folder to get some ideas

students/ahwall/Lesson05Homework/MailRoomPt3.py
```diff
@@ -0,0 +1,151 @@
...
+                else:
+                    Don_Val = ()
+                    Don_Val = (ty_letter[:], vallist[:])
+                    donors.append(Don_Val)
+                    print()
+                    print(ty_letter, "and their donation have been added to \
+                    the Donor Report...Creating 'Thank you email'")
+                    i = 1
+                    One_letter(ty_letter, DonationVal)
+                    mainloop()
```
HarryMaher: Note that by calling mainloop() you're building the stack, so you might want to avoid doing this

students/ahwall/Lesson05Homework/MailRoomPt3.py
```diff
@@ -0,0 +1,151 @@
...
+
+
+def report(donors):
+    Repheader = ('Donor Name', 'Total Given', 'Num Gifts', 'Average Gifts')
+    headerspace = '{:20}| {:^14}| {:^10}| {:^14}'
+    ReportCats = headerspace.format(*Repheader)
+    print('\n', ReportCats)
+    print('-' * 63)
+    # for loop to print report
+    [print('{:20}| ${:^13,.02f}| {:^10}| ${:^14,.02f}'.format(item[0],
```
HarryMaher: Nice list comprehension!  I think the ideal case would put the print separate from the list comp, so you only have to call print() once, but this works!

students/ahwall/Lesson05Homework/MailRoomPt3.py
```diff
@@ -0,0 +1,151 @@
...
+        elif response == "2":
+            report(donors)
+        elif response == "3":
+            All_letters(donordict)
+        elif response == "4":
+            quit()
+
+
+if __name__ == "__name__":
+    mainloop()
```
HarryMaher: Works for sure, but some refactoring would help with efficiency. Python really doesn't use "i" or "j" or whatever to loop through code

PR #67, opened 2018-04-29
Mailroom 2 and kata fourteenv (left open, as lesson 7 commits are already piling up here)

weepler:
Please see the following pull request containing Mailroom 2 and the trigrams exercise in the lesson 4 and lesson 5 folder. 

Best regards, 

Wayne


students/Wayne/Assignments/Lesson5/Mailroom2.py
```diff
@@ -0,0 +1,256 @@
...
+
+# ----------------------------------------------------------------------------
+#
+# The following function adds a new donation amount into the database
+#
+# ----------------------------------------------------------------------------
+
+
+def accept_donation(name):
+    while True:
```
hosungs: Why loop here? You are not repeating at all. You are either returning or breaking out of the loop, so this loop doesn't seem necessary.

students/Wayne/Assignments/Lesson6/test_Mailroom.py
```diff
@@ -0,0 +1,16 @@
...
+"""
+
+
+import Mailroom2 as MR
+
+
+def test_accept_donation():
+    amount = 1400
+    MR.accept_donation( amount)
+    assert MR.donor_db == {[1400]}
```
hosungs: This test seems misleading. Your accept_donation() expects a name, not an amount. Tests should reflect the spec of the code that's tested.

hosungs:
Leaving this open, as commits for lesson 7 are already piling up on this PR...

PR #68, opened 2018-04-30
Osiddiquee up to lesson06 (left open because no mailroom test code is committed yet)

osiddiquee:
commits until lesson 06


students/Osiddiquee/lesson05/mailroom3.py
```diff
@@ -0,0 +1,129 @@
...
+                print('Enter an amount')
+                amount = int(input('>> '))
+                try:
+                    float(amount)
+                    for (donor, amounts) in donors:
+                        if name == donor:
+                            amounts.append(amount)
+                except ValueError:
+                    print('Please enter a valid donation')
+                    return
```
hosungs: Did you mean `continue` here? `return` will break out of the loop, so no more input will be asked. Note that `continue` is not needed as well, because there's no statements that follow this line in the loop. However, it's good to have `continue` here, just in case some other code might be added outside this whole nested if structure (but inside the while loop).

Overall, it's not too desirable to have too deep nested structure. Try to separate some part and create a function for it (we call this "refactoring"). The amount input acceptance/validation part is a good choice for such refactoring.

hosungs:
I don't see any mailroom test code, so I just updated your Lesson 5 grade, but leaving this open for your Lesson 6 submission (mailroom test code).

students/Osiddiquee/lesson08/circle.py
```diff
@@ -0,0 +1,41 @@
...
+        return math.pi * self.radius**2
+
+    def __str__(self):
+        return f'Circle with radius: {self.radius}'
+
+    def __repr__(self):
+        return f'Circle({self.radius})'
+
+    def __add__(self, other):
+        return f'Circle({self.radius + other.radius})'
```
hosungs: This is not correct. A Circle object should be returned as the result of this addition, not a string representation. Please fix.

students/Osiddiquee/lesson08/circle.py
```diff
@@ -0,0 +1,41 @@
...
+
+    def __mul__(self, number):
+        return f'Circle({self.radius * number})'
```
hosungs: Same issue.

students/Osiddiquee/lesson08/test_circle.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+def test_str():
+    c = Circle(4)
+    print(c)
+
+
+def test_repr():
+    c = Circle(4)
+    repr(c)
```
hosungs: Should have an assert on this result.

students/Osiddiquee/lesson08/test_circle.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+def test_numeric():
+    c1 = Circle(2)
+    c2 = Circle(4)
+
+    assert c1 + c2 == 'Circle(6)'
+    c2 * 3 == 'Circle(12)'
+    3 * c2 == 'Circle(12)'
```
hosungs: Not the right tests. Results of these operations should be Circle objects, not strings. Please fix.

PR #70, opened 2018-04-30
Andy Kwon - Lesson06 Pull Request

KwonAndy:
PR for the mailroom unit tests. Includes updated mailroom code to allow for better testing.

Thank you,

Andy

### DIFF
--- a/students/andykwon/lesson06/mailroom_pt4.py
+++ b/students/andykwon/lesson06/mailroom_pt4.py
@@ -50,19 +50,17 @@ def check_if_number(response):
         return False
 
 
-
 def add_donation():
 
     print("Please enter the name of the donor:\n")
     donor_name = input(">>")
 
     donation_amount = input_loop_for_add_donation()
-    
+
     # while True:
     #     print("Please enter the amount " + donor_name + " donated:\n")
     #     entered_amount = input(">>")
     #     donation_amount = check_if_number(entered_amount)
-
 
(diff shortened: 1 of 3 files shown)


students/andykwon/lesson06/mailroom_pt4.py
```diff
@@ -0,0 +1,201 @@
...
+    2) Report
+    3) Send etters to everyone
+    4) Quit\n"""
+
+    run_loop(options_dict, menu_string)
+
+
+if __name__ == "__main__":
+    print("...initiating...")
+    mainloop()
```
hosungs: Overall, looks good. However, mainloop() calling runloop(), which may call quit_men() that calls runloop(), ... It's not an arbitrary recursion, just one level, but it looks a bit confusing. I see your point to reuse run_loop() code in multiple places, though.

students/andykwon/lesson06/test_mailroom_pt4.py
```diff
@@ -0,0 +1,98 @@
...
+    """
+    Test to see if entering a numerical value returns a float of that number
+    """
+    user_input = "100.00"
+    assert mailroom.check_if_number(user_input) == float(user_input)
+
+
+def test_return_to_menu():
+    result = mailroom.return_to_menu()
+    assert result is True
```
hosungs: Good tests!

PR #75, opened 2018-05-02
Unit Testing and Altered MainRoomPt3

ahwall78:


students/ahwall/Lesson06Homework/Lesson06HW_Git/MailRoomPt3.py
```diff
@@ -0,0 +1,167 @@
...
+\tSincerely,\n\n \tThe Python Project\n".format(donor, donation)  # Works
+
+
+def All_letters(donordict):
+    donorvallist = list(donordict.values())
+    for i in range(len(donorvallist[0])):
+        Donor, Donation = donordict['Donor'][i], donordict['Donation'][i]
+        Donation = int(Donation)
+        # could not figure out a way to unpack the dictionary (like Mr. Barker
+        # suggested in the question) with multiple values per key, since you
```
HarryMaher: Ohh yeah, this'll require you to change your dictionary structure to something like:
```
donor_dict{
"Donor 1": [21.45, 23.64, 2.67]
"Donor 2": [2.45, 23.64, 2.67]
}
```

students/ahwall/Lesson06Homework/Lesson06HW_Git/MailRoomPt3.py
```diff
@@ -0,0 +1,167 @@
...
+        format(item[0], sum(item[1]), len(item[1]), (sum(item[1]) /
+        len(item[1]))))
+        print(don_report)
+    return don_report
+    #don_report = [print('{:20}| ${:^13,.02f}| {:^10}| ${:^14,.02f}'.format(item[0],
+    #sum(item[1]), len(item[1]), (sum(item[1]) / len(item[1]))))
+    #for item in donors]
+    #  return don_report
+    #  mainloop()
+    # could not get it to pass test_All_letters with mainloop() still active
```
HarryMaher: Oh yeah, and calling to mainloop() builds up the stack, so you don't really want to do that, so yay for test-driven-development

students/ahwall/Lesson06Homework/Lesson06HW_Git/MailRoomPt3.py
```diff
@@ -0,0 +1,167 @@
...
+        if response not in ("1", "2", "3", "4"):
+            print("***Not a Valid response  -- type 1,2,3***\n")
+            continue
+        elif response == "1":
+            thank_you()
+        elif response == "2":
+            report(donors)
+        elif response == "3":
+            All_letters(donordict)
+        elif response == "4":
```
HarryMaher: I think I commented this in a past version, but dicts make much better switch statements in python. 

https://stackoverflow.com/questions/21962763/using-a-dictionary-as-a-switch-statement-in-python

students/ahwall/Lesson06Homework/Lesson06HW_Git/MailRoomPt3.py
```diff
@@ -0,0 +1,167 @@
...
+    else:
+        i = 0
+        while i < 1:
+            for item in donors:
+                DonorInList = False
+                if item[0].count(ty_letter) == 1:
+                    DonorInList = True  # works
+                    break
+            if DonorInList is True:
+                j = 0
```
HarryMaher: I don't know if I already commented on this, but man this is complicated. If you had time to refactor it'd be great to break this up and simplify. In Python for loops pretty much can always take the place of using i = 0 and i += 1.

students/ahwall/Lesson06Homework/Lesson06HW_Git/MailRoomPt3.py
```diff
@@ -0,0 +1,167 @@
...
+                for item in donors:
+                    if ty_letter == item[0]:  # Works
```
HarryMaher: Also, using dicts to their true potential would simplify this a lot. Here you're searching through each item in the dictionary. But with a dictionary we can use the key to immediately find the location of the value(s) we want to access

PR #76, opened 2018-05-03
commit mailroom part 1

tammytdo:


students/tammyd/lesson03/mailroom_pt1.py
```diff
@@ -0,0 +1,114 @@
...
+
+"""
+
+import sys
+
+donor_db = [("Britney Spears", [50.99, 100.99, 2000.99]),
+            ("Christina Aguilera", [50.99, 100.99, 2000.99]),
+            ("Mandy Moore", [50.99, 100.99, 2000.99]),
+            ("Jessica Simpson", [50.99, 100.99, 2000.99]),
+            ("Queen Bey", [50.99, 100.99, 2000.99]),
```
HarryMaher: Queen Bey! 👍

students/tammyd/lesson03/mailroom_pt1.py
```diff
@@ -0,0 +1,114 @@
...
+    print("Welcome to the Mailroom \n")
+
+def main_menu():
+    print("Main Menu:")
+    print(
+      "T - Send a thank you\n"
+      "R - Create a report\n"
+      "Q - Quit the program"
+      )
+    menu_selection = input("\nChoose a letter:  ")
```
HarryMaher: We probably want something like:
```
while True:
```
around our menu, so that  it keeps running without each function needing to call main_menu()

students/tammyd/lesson03/mailroom_pt1.py
```diff
@@ -0,0 +1,114 @@
...
+    if menu_selection.upper().strip() == "T":
+        thank_you()
+    elif menu_selection.upper().strip() == "R":
+        create_report()
+    elif menu_selection.upper().strip() == "Q":
```
HarryMaher: I like that you used .upper().strip()! A slightly more efficient way would be to put those methods on menu_selection before comparing them. Like:
```
menu_selection = input("select stuff!")
menu_selection = menu_selection.upper().strip()
if menu_selection == "T":
#etc. etc.
```

Also, in the future we can use dicts instead of a bunch of "elif"s.

students/tammyd/lesson03/mailroom_pt1.py
```diff
@@ -0,0 +1,114 @@
...
+
+    while True:
+        donor_amount = input("Enter amount of donation amount (or 'menu' to exit): ").strip()
+        if donor_amount.lower() == "menu":
+            return main_menu()
+        else:
+            break
+    #I borrowed this from the example solution. Could you please explain None to me? 
+    donor = find_donor(donor_name)
+    if donor is None:
```
HarryMaher:
```
donor = find_donor(donor_name)
# the above returns None if no donor is found 
if donor is None:  # if no donor is found in the data structure
donor = (donor_name, []) # then put one in.
```

students/tammyd/lesson03/mailroom_pt1.py
```diff
@@ -0,0 +1,114 @@
...
+        donor = (donor_name, [])
+        donor_db.append(donor_name)
+
+    donor[1].append(donor_amount)
```
HarryMaher: donor_amount here is still a string, so it's going to break if we try to do math in it, like we do in the report function. I would change it to:
```
donor[1].append(float(donor_amount)) #float changes your string to a floating point number, which we can do math with
```

students/tammyd/lesson03/mailroom_pt1.py
```diff
@@ -0,0 +1,114 @@
...
+def find_donor(donor_name):
+    for donor in donor_db:
+        if donor_name.strip().lower() == donor[0].lower():
+            return donor
+    return None
+
+
+def ty_letter(donor_name):
+    print("\nMenu: Send a Thank You, Letter")
+    #How can I input the donation amount?? Why does donor_amount not work here?
```
HarryMaher: donor_amount only exists in thank_you(), so this function,  ty_letter(), doesn't see it. You can pass it in as an argument here if you add a parameter for donor_amount. i.e. change your function definition to:
```
def ty_letter(donor_name, donor_amount):
```

PR #79, opened 2018-05-05
first commit mailroom pt 2

tammytdo:


students/tammyd/lesson05/mailroom_pt3.py
```diff
@@ -0,0 +1,171 @@
...
+            #     thank_you()
+            # elif menu_selection == "L":
+            #     letters_to_everyone()
+            # elif menu_selection == "R":
+            #     create_report()
+            # elif menu_selection == "Q":
+            #     quit_program()
+
+        #why is my exception not being raised here when I make an invalid entry?
+        except (KeyError):
```
HarryMaher: So you're catching the keyerrors so the exception isn't being raised. I would delete the return on line 61 so that this can re-run the menu

students/tammyd/lesson05/mailroom_pt3.py
```diff
@@ -0,0 +1,171 @@
...
+
+def donation_amount(donor_name):
+    while True:
+        donor_amount = input("Enter amount of donation amount (or 'menu' to exit): ").strip()
+        if donor_amount.lower() == "menu":
+            return main_menu()
+            #what is the difference between these three?...
+               #return
+               #return main_menu
+               #return main_menu()
```
HarryMaher: On its own, "return" returns nothing, which effectively brings you back to the main menu and that's what you should do to prevent building up the call stack. return main_menu() builds the call stack by calling main menu when you don't need to. and return main_menu returns a function, main menu, without calling it, so that's not what you want.

students/tammyd/lesson05/mailroom_pt3.py
```diff
@@ -0,0 +1,171 @@
...
+    print("-" * 61)
+    for row in report_rows:
+    #need to correct my formatting of the $ symbol
+        print("{:20s} ${:2.2f} {:5d} ${:14.2f}".format(*row))
+    print("-"*61)
+    print("\n")
+    return main_menu()
+ 
+
+def quit_program():
```
HarryMaher: Looks good overall!

PR #82, opened 2018-05-07
 Lesson07: HTML Render - kristianjf

kristianjf:
Would like to receive feedback on HTML Render Lesson07.

This lesson took 2 days.

List of Challenges:

- test_html_render.py: test_render_element
> Attempted dictionary comprehension with referencing dictionary being created. Found out this does not work. Used Counter method from collections instead to verify number of <html> tags == 1

- html_render.py: class Element - def add_newlines
> created function for handling newlines. Good lessons learned. Found out that str is an object so have to test string when attempting to catch objects that are not strings. Any feedback on whether creating this function (add_newlines) was an effective approach is appreciated.

- html_render.py: class Element - def render
> Handling indentation took some time. Passing objects between classes was a bit complex. My solution was to initialize 'ind' as a class attribute and use the render loop to pass incremented kwargs['indent'] to inherited render functions.

Lessons Learned:
- Unit Testing is a life-saver. Nearly impossible to do this exercise without it.
- Appreciate a good challenging lesson with a lot of good examples to reinforce lectures/notes.


students/kristianjf/lesson07/html_render.py
```diff
@@ -0,0 +1,156 @@
+#!/usr/bin/env python3
+
+"""
+A class-based system for rendering html.
+"""
+
+
+# This is the framework for the base class
+class Element:
+    'Base Class for HTML Render'
```
hosungs: Use docstring comment:
```
"""Base Class ..."""
```

students/kristianjf/lesson07/html_render.py
```diff
@@ -0,0 +1,156 @@
...
+    tag = 'html'
+    ind = 0
+
+    def __init__(self, content=None, **kwargs):
+        if content is None:
+            self.content = []
+        if content != None:
```
hosungs: Should be just `else:`

students/kristianjf/lesson07/html_render.py
```diff
@@ -0,0 +1,156 @@
...
+            self.content = [content]
+        if kwargs:
+            self.attrs = kwargs
+        else:
+            self.attrs = None
```
hosungs: Probably `self.attrs = {}` ?

students/kristianjf/lesson08/circle.py
```diff
@@ -0,0 +1,41 @@
...
+        self = cls(int(diameter_value/2))
+        return self
+    def __str__(self):
+        return f'Circle with radius: {self.radius}'
+    def __repr__(self):
+        return f'Circle({self.radius})'
+    def __add__(self, other):
+        return Circle(self.radius + other.radius)
+    def __mul__(self, other):
+        return Circle(self.radius * other.radius)
```
hosungs: This is not the intended use case. Note in the example for the multiplication operation that the `other` is an integer, not a circle.

kristianjf:
@HarryMaher / @hosungs  , my sincere apologies. I did not notice the comments on previous assignments until today. I will review all the comments and use them in the future. Greatly appreciate the time you took here to provide feedback. A bit disappointed in myself I did not notice this sooner.

PR #83, opened 2018-05-07
assignment for lesson06 & 08, still working on lesson07

rachelz11:


students/Ruohan/lesson07/html_render.py
```diff
@@ -0,0 +1,77 @@
...
+
+"""
+A class-based system for rendering html.
+"""
+
+
+# This is the framework for the base class
+class Element:
+    tag = 'html'
+    def __init__(self, content=None, style="line-height:200%"):
```
hosungs: I think we went over this last night together in person, but this PR doesn't seem to include your latest changes. An element can have any number of attributes, so use `**kwargs`, save the dict properly and render it in the opening tag.

students/Ruohan/lesson07/html_render.py
```diff
@@ -0,0 +1,77 @@
...
+        for line in self.content:
+            if isinstance(line, str):
+                out_file.write(line)
+                out_file.write('\n')
+            else:
+                line.render(out_file)
+        out_file.write(f'</{self.tag}>\n')
+
+
+class style(Element):
```
hosungs: `style` is not an HTML element, so no need for this.

hosungs:
Oops, there are merge conflicts and this PR can't be merged. Not sure if this is going to work, but please try this and let me know. If it doesn't work, do contact us in class or at office hours and get these conflicts resolved.

```
$ git pull -X theirs upstream master
$ git push origin master
```

PR #84, opened 2018-05-08
Lesson07 assignment submission

rdrovdahl:
html render assignment

### DIFF
--- a/students/drovdahl/lesson07/html_render.py
+++ b/students/drovdahl/lesson07/html_render.py
@@ -1,0 +1,161 @@
+#!/usr/bin/env python3
+
+"""
+A class-based system for rendering html.
+Ryan Drovdahl's version
+"""
+
+
+# This is the framework for the base class
+class Element(object):
+
+    # 'html' is the default tag that is only used for testing
+    tag = 'html'
+    close_tag_line = True
+    link_tag = False
+    meta_tag = False
+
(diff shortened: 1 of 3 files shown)


students/drovdahl/lesson07/html_render.py
```diff
@@ -0,0 +1,182 @@
...
+# This is the framework for the base class
+class Element(object):
+
+    # 'html' is the default tag that is only used for testing
+    tag = 'html'
+    close_tag_line = True
+    link_tag = False
+    meta_tag = False
+
+    def __init__(self, content=None, style=None, id=None, indent=4):
```
hosungs: `style` and `id` are not the only attributes that may be specified in an HTML element. Therefore, you should have received `**kwargs` to accommodate arbitrary attributes.

students/drovdahl/lesson07/html_render.py
```diff
@@ -0,0 +1,182 @@
...
+        self.indent = indent
+
+    def append(self, new_content):
+        self.content.append(new_content)
+
+    def open_tag(self):
+        if self.id and self.style:
+            return f'<{self.tag} id="{self.id}" style="{self.style}">'
+        elif self.style:
+            return f'<{self.tag} style="{self.style}">'
```
hosungs: And this code should become a loop over the dict, not this simple if statements...

PR #86, opened 2018-05-09
pull request for cirlces.py and html render 

weepler:
Please see the following pull requests for the following assignments: 

- Circle.py 

- html_render.py


hosungs:
@weepler -- Your html_render.py looks OK, so I'm marking your html_render assignment as complete.

However, I can't find your circle.py in this PR. Besides, there are many other files from other students (or from instructor's directories) that are added here, which is very weird. All in all, there are merge conflicts so I can't merge this PR without resolving the conflicts first. We'll do that in class tomorrow.

PR #88, opened 2018-05-10
drovdahl - added circle exercise and test code.

rdrovdahl:
Circle assignment ready for review.


students/drovdahl/lesson08/circle.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+class Circle(object):
+    def __init__(self, radius):
+        self._radius = radius
+        self.diameter = radius * 2
+
+    @property
+    def radius(self):
+        return self.diameter / 2
```
hosungs: Why not just `return self._radius`?

students/drovdahl/lesson08/circle.py
```diff
@@ -0,0 +1,74 @@
...
+
+    @property
+    def diameter(self):
+        return self._radius * 2
```
hosungs: Why not just `return self.diameter`?

students/drovdahl/lesson08/circle.py
```diff
@@ -0,0 +1,74 @@
...
+
+    @diameter.setter
+    def diameter(self, val):
+        self._radius = val / 2
```
hosungs: Since you keep track of diameter separately (which you probably shouldn't, because the diameter can always be computed from the radius), you need to update `self.diameter` as well (`self.diameter = val`). Otherwise, subsequent calls to various methods will give incorrect values.

students/drovdahl/lesson07/html_render.py
```diff
@@ -0,0 +1,182 @@
...
+# This is the framework for the base class
+class Element(object):
+
+    # 'html' is the default tag that is only used for testing
+    tag = 'html'
+    close_tag_line = True
+    link_tag = False
+    meta_tag = False
+
+    def __init__(self, content=None, style=None, id=None, indent=4):
```
hosungs: Still not as expected. Should receive `**kwargs` instead of fixed `style=..., id=...`. See https://github.com/UWPCE-PythonCert-ClassRepos/Wi2018-Classroom/blob/master/solutions/Session07/step_8/html_render.py

students/drovdahl/lesson08/circle.py
PythonCHB: Comment from the peanut gallery:

The point of this exercise is that you only want to store either radius or diameter, but not both -- and calculate one from the other with a property when needed.

and in this case, you have defined a "diameter" property -- and then written over it with a "diameter" instance attribute. So the property will not be used at all.

Did this code work? In the sense of being able to change either radius or diameter and then have both still be correct?

students/drovdahl/lesson08/test_circle.py
```diff
@@ -0,0 +1,92 @@
...
+    c = Circle(10)
+    assert c.radius == 10
+
+
+def test_diameter():
+    c = Circle(2)
+    assert c.diameter == 4
+
+
+def test_diameter_change():
```
PythonCHB: did this test pass??

students/drovdahl/lesson08/circle.py
rdrovdahl: The code worked.  I'll refactor it and try it out the way you're suggesting as well.  thanks for the feedback!

students/drovdahl/lesson08/test_circle.py
rdrovdahl: Yes, all tests passed.

students/drovdahl/lesson07/html_render.py
rdrovdahl: I haven't had time to refactor this yet.  I have every intention of revisiting this though.  I spent a number of hours trying to make it work with **kwargs the first time and eventually gave up.  Given enough time however, I'd like to try and fix it.

students/drovdahl/lesson08/circle.py
rdrovdahl: OK, I 'think' I understand what you're saying here.  If you could give me an example of how a subsequent call to another method would fail given this code, that would be really helpful.

students/drovdahl/lesson08/circle.py
rdrovdahl: Not sure why I did that...  Probably because it was 1AM . :)

students/drovdahl/lesson08/circle.py
PythonCHB: I see:

You have a radius property that is computing from diameter. so re-setting the diameter works.

but the self.diameter is over-riding the diameter property, so it isn't doing anything, and self._radius isn't used as a result.

I think you would find that you can't re-set the radius with this code -- there must not have been a test for that -- so hard to get comprehensive tests!

rdrovdahl: You're ABSOLUTELY correct!  I just tried to reset the radius and failed...

Still not sure I completely understand this entirely but will spend some more cycles on this and try and get a handle on it.

students/drovdahl/lesson08/circle.py
PythonCHB: looking now, if you comment out the 

``self.diameter = ``

line it will all work -- except being able to set the radius.

So I'd change self._radius to plain old self.radius, and get rid of the radius property.

students/drovdahl/lesson08/circle.py
PythonCHB: radius was a property with only a getter, and no setter, so you can't set it.

rdrovdahl: Ah...  makes sense.  I might have missed where the assignment asked for a radius setter.

PythonCHB: it didn't -- it probably didn't specify, but it was expecting that radius would be a plain old attribute. But if you DO make it a property, you probaby want to provide a setter.

we should add that to the test suite.

rdrovdahl: Oh, I see!  So, is it pretty much standard practice to configure both a setter and getter for any property you make?

students/drovdahl/lesson08/circle.py
rdrovdahl: This is addressed in my next PR

students/drovdahl/lesson08/circle.py
rdrovdahl: This is addressed in my next PR

students/drovdahl/lesson08/circle.py
rdrovdahl: This is addressed in my next PR

PR #89, opened 2018-05-10
commit mailroom pt 4

tammytdo:


students/tammyd/lesson06/test_mailroom_pt4.py
```diff
@@ -0,0 +1,64 @@
...
+   donor = mailroom.find_donor("queeN bey")
+   assert donor[0] == "Queen Bey" 
+#   donor = mailroom.find_donor("mandy MOOre")
+#   assert donor[0] == "Mandy Moore"
+
+#Why does this pass even when I test "qenn bey" instead of "Queen Bey"?
+def test_find_donor_not():
+    "test one that's not there"
+    donor = mailroom.find_donor("qenn bey")
+    assert donor is None
```
HarryMaher: This will pass if None is None. And since find_donor won't find "qenn bey" it'll return None. Then you'll assert donor, which means none is none, so that'll pass. It won't pass if you can find the donor. It's exactly what it looks like it should do.

HarryMaher:
Nice work! It looks like you have an understanding of some of what you're supposed to test. Ideally tests would be more thorough, and I couldn't get a lot of them to run, but it looks like you've put in a good faith effort, so keep up the progress!

PR #91, opened 2018-05-10
addressed numerous cleanup items on circle.py

rdrovdahl:
I believe I've fixed everything that was pointed out as being wrong earlier.

### DIFF
--- a/students/drovdahl/lesson08/circle.py
+++ b/students/drovdahl/lesson08/circle.py
@@ -18,24 +18,20 @@ import math
 
 class Circle(object):
     def __init__(self, radius):
-        self._radius = radius
+        self.radius = radius
         self.diameter = radius * 2
 
     @property
-    def radius(self):
-        return self.diameter / 2
-
-    @property
     def diameter(self):
-        return self._radius * 2
+        return self.radius * 2
 
     @diameter.setter
     def diameter(self, val):
-        self._radius = val / 2
+        self.radius = val / 2
 
     @property
     def area(self):
-        return math.pi * self._radius ** 2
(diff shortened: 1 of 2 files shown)


students/drovdahl/lesson08/circle.py
```diff
@@ -18,53 +18,49 @@
 
 class Circle(object):
     def __init__(self, radius):
-        self._radius = radius
+        self.radius = radius
         self.diameter = radius * 2
```
HarryMaher: You can also delete this "self.diameter = radius * 2" on line 22, because you defined diameter as a @property below and it already returns radius * 2.

students/drovdahl/lesson08/circle.py
```diff
@@ -18,53 +18,49 @@
...
-        return self._radius > other._radius
+        return self.radius > other.radius
 
     def __lt__(self, other):
-        return self._radius < other._radius
+        return self.radius < other.radius
 
     def __eq__(self, other):
-        return self._radius == other._radius
+        return self.radius == other.radius
```
HarryMaher: Looks good to me!

students/drovdahl/lesson08/test_circle.py
```diff
@@ -1,8 +1,8 @@
 #! /usr/local/bin/python3
 '''
-test code for circle.py
+test code for circle2.py
 '''
-from circle import *
+from circle2 import *
```
HarryMaher: Nice work! Don't forget to change your circle2.py back to circle! But I'm just being annoying. Goood job!

PR #92, opened 2018-05-11
commit functional mailroom. from class

tammytdo:


students/tammyd/lesson10/mailroom2_working.py
```diff
@@ -0,0 +1,277 @@
...
+
+    @property
+    def name(self):
+        return self._name
+
+    @property
+    def donations(self):
+        return self._donations
+
+    def add_donations(self, donations):
```
PythonCHB: this looks like untested code --- it has "donations" as a parameter, but is adding "donation" to the list.

you probably want the singular form in both cases.

students/tammyd/lesson10/mailroom2_working.py
```diff
@@ -0,0 +1,277 @@
...
+        self._donations.append(donation)
+
+    @property
+    def total_donations(self):
+        return reduce(lambda a,x: a+t, self._donations, 0)
```
PythonCHB: untested? lambda a,x then a+t ??

of course, this is just the sum() function -- but good as a way to practice reduce and lambda.

students/tammyd/lesson10/mailroom2_working.py
```diff
@@ -0,0 +1,277 @@
...
+        return "{}: {}".format(self._name, self._donations)
+
+class DonorDB:
+    def __init__(self):
+        self._donors = {}
+
+    def add_donor(sef, donor):
+        self._donors[donor.name.lower()] = donor
+
+    def get_total_from_donor(self, donor_name):
```
PythonCHB: I'd be inclined to leave this in the donor class.

students/tammyd/lesson10/mailroom2_working.py
```diff
@@ -0,0 +1,277 @@
...
+        self._donors[donor_name.lower()].total_donations
+
+    def get_donor(self, donor_name):
+        return self._donors[donor_name.lower()] ###stuff here
+
+    @property
+    def num_donors(self):
+        return str(self._donors)
```
PythonCHB: num_donors returns a stringified dict of the donor objects???

students/tammyd/lesson10/mailroom2_working.py
```diff
@@ -0,0 +1,277 @@
...
+
+    def __repr__(self):
+        return #whaaaax
+
+def get_donor_db_oo():
+    db = DonorDB()
+    raw_data = get_donor_db()
+
+    for k, v in raw_data.items():
+        donor = Donor(k)
```
PythonCHB: Donor can take alist of donations to initialize, so this can be done in one fell swoop ;-)

students/tammyd/lesson10/mailroom2_working.py
```diff
@@ -0,0 +1,277 @@
...
+        for donation in v:
+            donor.add_donation(donation)
+
+    return db
+
+
+
+
```
PythonCHB: now to move all of these into the classes -- if it acts on one donor, into the Donor class, if it acts on the collection, into the DonorDB class.

Then, presto! and OO mailroom :-)

PR #93, opened 2018-05-11
Andy K - html render PR (and maybe mailroom oo and functional)

KwonAndy:


students/andykwon/lesson07/html_render.py
```diff
@@ -77,17 +92,64 @@ class Title(OneLineTag):
 
 class SelfClosingTag(Element):
 
-    def render(self, outfile)
-        pass
+    def append(self, *args, **kwargs):
+        raise TypeError
```
PythonCHB: Good to add a message to the error:

`raise TypeError("you cannot add content to a Self Closing Tag")`

students/andykwon/lesson07/html_render.py
```diff
@@ -77,17 +92,64 @@ class Title(OneLineTag):
...
+
+    def open_tag(self,):
+
+        line = ''
+        opening = f"<{self.tag}"
+
+        if self.attr != {}:
```
PythonCHB: an empty dict is "falsey" so this can be:

`if self.attr`

students/andykwon/lesson07/html_render.py
```diff
@@ -77,17 +92,64 @@ class Title(OneLineTag):
...
+            for key, value in self.attr.items():
```
PythonCHB: this is ALMOST the same code as in the base class -- could you refactor so as not to repeat yourself?

students/andykwon/lesson07/html_render.py
```diff
@@ -77,17 +92,64 @@ class Title(OneLineTag):
...
     tag = 'br'
 
+class A(OneLineTag):
+
+    tag = "a"
+
+    def __init__(self, link, content=None, **kwargs):
+        kwargs['href'] = link
+        # Element.__init__(self, content, **kwargs)
+        super().__init__(content, **kwargs)
```
PythonCHB: nice!

PR #96, opened 2018-05-12
lesson09 oo mailroom

kristianjf:
OO Mailroom Assignment plus unit testing

### DIFF
--- a/students/kristianjf/lesson09/mailroom_5.py
+++ b/students/kristianjf/lesson09/mailroom_5.py
@@ -1,0 +1,221 @@
+#!/usr/bin/env python3
+import sys
+'''
+Goal:
+You work in the mail room at a local charity. Part of your job is to write \
+incredibly boring, repetitive emails thanking your donors for their generous \
+gifts. You are tired of doing this over and over again, so you’ve decided to \
+let Python help you out of a jam and do your work for you.
+'''
+'''
+It should have a data structure that holds a list of your donors and a history of \
+the amounts they have donated. This structure should be populated at first with at \
+least five donors, with between 1 and 3 donations each.
+'''
+# self.DATA_STRUCTURE = [('Jeff Bezos', [200, 500, 1000]), \
+#                     ('Bill Gates', [1000]), \
+#                     ('Steve Jobs', [20, 50, 100])]
+# self.DATA_STRUCTURE_DICTS = [{'name': name, 'donation': donation} for name, donation in self.DATA_STRUCTURE]
+# # self.DATA_STRUCTURE_DICT = [ \
+# #                         {'name': 'Jeff Bezos', 'donation': [200, 500, 1000]},\
+# #                         {'name': 'Bill Gates', 'donation': [1000]},\
+# #                         {'name': 'Steve Jobs', 'donation': [20, 50, 100]}\
+# #                         ]
+
+class Donor:
+    def __init__(self, name, donations):
+        self._name = name
+        self._donations = [] if donations is None else donations
+    @property
+    def name(self):
+        return self._name
+    @property
+    def donations(self):
+        return self._donations
+    def donate(self, new_donation):
+        self.donations.append(new_donation)
+    def metrics(self):
+        self.num_d = len(self.donations)
+        self.total_d = sum(self.donations)
+        self.avg_d = self.total_d / self.num_d
+
+class DonorsHandle:
+    DATA_STRUCTURE = [] #[('',[])]
+    DATA_STRUCTURE_DICTS = [] #[{'name': None, 'donation': None}]
+    def __init__(self):
+        pass
+    def add_donor(self, donor_object):
+        self.DATA_STRUCTURE.append((donor_object.name, donor_object.donations))
+        self.DATA_STRUCTURE_DICTS.append({'name': donor_object.name, \
+                                    'donation': donor_object.donations})
+
+    def thank_you(self, full_name='', donation_amount=int()):
+        '''
+        If the user (you) selects ‘Send a Thank You’, prompt for a Full Name.
+            If the user types a name not in the list, add that name to the data structure and use it.
+            If the user types a name in the list, use it.
+        Once a name has been selected, prompt for a donation amount.
+            Turn the amount into a number – it is OK at this point for the program to crash if \
+            someone types a bogus amount.
+        Once an amount has been given, add that amount to the donation history of the selected user.
+        Finally, use string formatting to compose an email thanking the donor for their generous \
+        donation. Print the email to the terminal and return to the original prompt.
+        '''
+        '''
+        Collect, Format and Validate Full name
+        '''
+        #full_name = ''
+        while not full_name:
+            full_name_input = input('Please enter full name of the recipient: ').split()
+            full_name = self.name_response_valid(full_name_input)
+        '''
+        Process Full Name
+        '''
+        ds_names = [name for name, donation in self.DATA_STRUCTURE]
+
+        if full_name in ds_names:
+            full_name_index = ds_names.index(full_name)
+            print(f'Found {full_name} in data_structure')
+            while not donation_amount:
+                donation_amount = int(input('Please enter a donation amount: '))
+            self.DATA_STRUCTURE[full_name_index][1].append(donation_amount)
+            print(self.output_string().format(name=full_name, donation=donation_amount))
+        else:
+            while not donation_amount:
+                donation_amount = int(input('Please enter a donation amount: '))
+            self.DATA_STRUCTURE.append((full_name, [donation_amount]))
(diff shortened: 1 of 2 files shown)


students/kristianjf/lesson09/mailroom_5.py
```diff
@@ -0,0 +1,221 @@
...
+        self._donations = [] if donations is None else donations
+    @property
+    def name(self):
+        return self._name
+    @property
+    def donations(self):
+        return self._donations
+    def donate(self, new_donation):
+        self.donations.append(new_donation)
+    def metrics(self):
```
PythonCHB: Probably better to have this computed and returned when you need it, rather than stored in self -- as these value will be out of date when a donation is added.

still good to do it all at once, and a method can return three values:

``return num_d total_d, avg_d`` 

alternatively, you could make them properties:

```
@property
def num_donations:
    return len(self._donations)
```

students/kristianjf/lesson09/mailroom_5.py
```diff
@@ -0,0 +1,221 @@
...
+        self.num_d = len(self.donations)
+        self.total_d = sum(self.donations)
+        self.avg_d = self.total_d / self.num_d
+
+class DonorsHandle:
+    DATA_STRUCTURE = [] #[('',[])]
```
PythonCHB: putting these as class attributes means that each instance of the class will have the same copies.

You typically only wil have one instance, so that's not too bad, but if you do make ore than one, it could get ugly. If nothign else, creating instances for tests is helpful.

So better to make these instance attributes, but defining them in __init__:

```
def __init__(self):
    self.data_structure = []
    self.data_structure_dicts = []
```
Though a bit odd to have both of those -- but maybe I'll see why as I look at more of the code :-)

students/kristianjf/lesson09/mailroom_5.py
```diff
@@ -0,0 +1,221 @@
...
+    DATA_STRUCTURE_DICTS = [] #[{'name': None, 'donation': None}]
+    def __init__(self):
+        pass
+    def add_donor(self, donor_object):
+        self.DATA_STRUCTURE.append((donor_object.name, donor_object.donations))
+        self.DATA_STRUCTURE_DICTS.append({'name': donor_object.name, \
```
PythonCHB: Why both? duplicate data!

But a bigger issue -- you have a Donor bject type -- you don't want to unpack it into a different data structure here, rather simply store the donor objects:

`self.donors.append(donor_object)`

or maybe make a nice key from the donor object ans store in a dic for fast lookup:

`self.donors[donor_object.name.lower().replace(' ','')] = donor_object`

or better yet:

`self.donors[donor_object.normalized_name] = donor_object`

where `normalized_name` is a property that computes a "normalized" name from the donor name.

This is all so that you can look up a name even if the capitalization, or spaces, or .. is different.

students/kristianjf/lesson09/mailroom_5.py
```diff
@@ -0,0 +1,221 @@
...
+        Once an amount has been given, add that amount to the donation history of the selected user.
+        Finally, use string formatting to compose an email thanking the donor for their generous \
+        donation. Print the email to the terminal and return to the original prompt.
+        '''
+        '''
+        Collect, Format and Validate Full name
+        '''
+        #full_name = ''
+        while not full_name:
+            full_name_input = input('Please enter full name of the recipient: ').split()
```
PythonCHB: this logic of the interface with the progra user should be separate from the class -- either stand alone functions or a UI class. That way you can test the functionality with no user interaction, and you could put a different UI on it -- maybe behind a web app?

students/kristianjf/lesson09/mailroom_5.py
```diff
@@ -0,0 +1,221 @@
...
+    def donations(self):
+        return self._donations
+    def donate(self, new_donation):
+        self.donations.append(new_donation)
+    def metrics(self):
+        self.num_d = len(self.donations)
+        self.total_d = sum(self.donations)
+        self.avg_d = self.total_d / self.num_d
+
+class DonorsHandle:
```
PythonCHB: DonorHandle? or DonorHandler?

or maybe DonorCollection, or .....

You want your names to reflect what the the object IS or does.

students/kristianjf/lesson09/mailroom_5.py
```diff
@@ -0,0 +1,221 @@
...
+            self.DATA_STRUCTURE[full_name_index][1].append(donation_amount)
+            print(self.output_string().format(name=full_name, donation=donation_amount))
+        else:
+            while not donation_amount:
+                donation_amount = int(input('Please enter a donation amount: '))
+            self.DATA_STRUCTURE.append((full_name, [donation_amount]))
+            print(self.output_string().format(name=full_name, donation=donation_amount))
+        return self.output_string().format(name=full_name, donation=donation_amount)
+
+    def list_donors(self):
```
PythonCHB: excelent -- stand alone functionality in the right class :-)

don't forget to keep your docstrings up to date ...

students/kristianjf/lesson09/mailroom_5.py
```diff
@@ -0,0 +1,221 @@
...
+        '''
+        If the user types ‘list’, show them a list of the donor names and re-prompt
+        '''
+        for entry in self.DATA_STRUCTURE:
+            print(entry[0])
+        return [entry[0] for entry in self.DATA_STRUCTURE]
+
+    def create_a_report(self):
```
PythonCHB: nicely done.

students/kristianjf/lesson09/mailroom_5.py
```diff
@@ -0,0 +1,221 @@
...
+            name, total_given, num_gifts, avg_gift = i[0], sum(i[1]), len(i[1]), sum(i[1]) / len(i[1])
+            format_string += format_data.format(name, total_given, num_gifts, avg_gift)
+        print(format_string)
+        return format_string
+
+    def key_index(self, item):
+        '''Sort by total sum of donation amount'''
+        return sum(item[1])
+
+    def send_letters(self):
```
PythonCHB: most of this logic should be in the Donor class, then in this class, you'd loop through each donor:

for entry in self.DATA_STRUCTURE_DICTS.values():
    entry.send_letter()

students/kristianjf/lesson09/mailroom_5.py
```diff
@@ -0,0 +1,221 @@
...
+        for entry in self.DATA_STRUCTURE_DICTS:
+            filename = '_'.join(entry.get('name').split())+'.txt'
+            filenames = []
+            with open(filename, 'w') as outfile:
+                outfile.write(self.output_string().format(**entry))
+                print(f'Writing: {filename}')
+                filenames.append(filename)
+        return filenames
+
+    def output_string(self):
```
PythonCHB: and this should be in the Donor class -- a Donor knows how to write a letter to its donor.

students/kristianjf/lesson09/mailroom_5.py
```diff
@@ -0,0 +1,221 @@
...
+        '''Output string for letters'''
+        format_string = 'Dear {name},\n\nThank you for your generous donation of ' \
+                        '${donation}. Please send us more money at your earliest ' \
+                        'convenience.'
+        return format_string
+
+    #Function to Process Menu Options
+    def menu(self, options_dict):
```
PythonCHB: again, better to keep the user interaction code outside the logic code.

students/kristianjf/lesson09/mailroom_5.py
```diff
@@ -0,0 +1,221 @@
...
+                full_name_cap += f'{name.capitalize()} '
+            full_name = full_name_cap.strip()
+        except AssertionError:
+            print('Invalid Name. Found non-alphabetic characters')
+        else:
+            return full_name
+
+def main():
+    'Create Donors'
+    jb = Donor('Jeff Bezos', [1, 5, 10])
```
PythonCHB: nice to have a function to create a populated DB, so it can be used in tests, too.

students/kristianjf/lesson09/test_mailroom_5.py
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env python3
+
+import mailroom_5 as m5
```
PythonCHB: good start on the tests -- but more to do!

kristianjf:
Thank you for all the feedback! I learned a lot through refactoring based on your comments. I was able to achieve the following:

- Removed list/dict data_structure and used objects completely
- Separated user menu (functions) and menu actions (methods)
- Fixed donor initialization (list or int accepted)

Action Items in https://github.com/UWPCE-PythonCert-ClassRepos/Sp2018-Accelerated/pull/114

PR #97, opened 2018-05-13
Lesson5: Mailroom version3 with try/except and list comprehension

meet29in:


students/meet_shah/lesson05/mailroom3.py
```diff
@@ -0,0 +1,121 @@
...
+        response = input(">>")
+        print("Your response was:", response)
+
+        ### added try/except clause
+        try:
+            myresponse = int(response)
+        except ValueError:
+            continue
+
+        try:
```
HarryMaher: These two are a little repetitive, and you only really need the 2nd one. I guess you could have done a different try/except in the thank_you() function

students/meet_shah/lesson05/mailroom3.py
```diff
@@ -0,0 +1,121 @@
...
+
+            ### Added list comprehension ###
+            [print(donor) for donor, _ in donors.items()]
+        elif new_response in ['q','quit']:
+            print("Returning to the main menu\n")
+            break
+        else:
+            donor = new_response
+            donation = input("Enter the amount you would like to donate: ")
+            donation = int(donation)
```
HarryMaher: Casting donation to float() would allow it to accept donations with cents too

students/meet_shah/lesson05/mailroom3.py
```diff
@@ -0,0 +1,121 @@
...
+        }
+
+def thank_you():
+    while True:
+        new_response = input("Enter a command. Your options are [list, <donor name>, q(quit),]>>")
+        if new_response == "list":
+            print("\nHere's a list of our donors:\n")
+
+            ### Added list comprehension ###
+            [print(donor) for donor, _ in donors.items()]
```
HarryMaher: Nice! Since you don't need the value. it'd be slightly more efficient to do
```
[print(donor) for donor in donors]
```
Because if you don't specify, it'll just give you the key, and some teacher once told me that ways more efficient than specifically telling it to give the donors.keys().

PR #98, opened 2018-05-13
Andy K - Commit mailroom_pt5.py and associated test. 

KwonAndy:
Object oriented version of mailroom pt 5.


students/andykwon/Lesson09/mailroom_pt5.py
```diff
@@ -0,0 +1,243 @@
...
+class Donations():
+
+    def __init__(self, donations=None):
+
+        if donations is None:
+            self.donations = {}
+        else:
+            self.donations = donations
+
+    def __str__(self):
```
HarryMaher: Nice touch with the str

students/andykwon/Lesson09/mailroom_pt5.py
```diff
@@ -0,0 +1,243 @@
...
+
+        for name in self.donations:
+            donors.append(name)
+
+        printable_donors = "\n".join(donors)
+
+        return printable_donors
+
+
+# -------------------------------------------------------------
```
HarryMaher: Ideally we wanted two classes -- one with donor-specific data, the other with the donor database, made up of of donors, but this works

students/andykwon/Lesson09/test_mailroom_pt5.py
```diff
@@ -0,0 +1,98 @@
...
+
+
+def thank_you():
+    """
+    for test_selection_happy() 
+    """
+    return "happy"
+
+
+def test_check_if_number_sad():
```
HarryMaher: Haha, nice test names. I'm fine with it as long as you have docstrings, which you do!

students/andykwon/Lesson09/test_mailroom_pt5.py
```diff
@@ -0,0 +1,98 @@
...
+
+def test_check_if_number_happy():
+    """
+    Test to see if entering a numerical value returns a float of that number
+    """
+    user_input = "100.00"
+    assert mailroom.check_if_number(user_input) == float(user_input)
+
+
+def test_return_to_menu():
```
HarryMaher: This seems like a fairly complete group of tests. Another possible idea is to test to make sure files have been written with their specified names and their content matches what's expected

PR #99, opened 2018-05-13
Lessons 06, 09, and 10 updated

DwightJay:
Forgot to add lesson08 for the circle exercise, I'll make another PR for it soon.  Lesson09 is complete but Lesson10 needs more work.  I should have another PR for that tonight, just wanted to add what I had so far.

### DIFF
--- a/students/jayjohnson/lesson06/mailroom_final.py
+++ b/students/jayjohnson/lesson06/mailroom_final.py
@@ -1,0 +1,257 @@
+#!/usr/bin/env python3
+'''
+Dwight Johnson - MailRoom - Final
+'''
+import sys
+from textwrap import dedent  # nifty utility!
+import math
+
+
+class MailRoom:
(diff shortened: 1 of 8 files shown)


students/jayjohnson/lesson06/mailroom_final.py
```diff
@@ -0,0 +1,257 @@
...
+    donor2.deposit(10000.00)
+    donor3.deposit(1000000.00)
+    donor3.deposit(50000.00)
+
+    donor_db = []
+
+    # add donors to database
+    donor_db.append(donor1)
+    donor_db.append(donor2)
+    donor_db.append(donor3)
```
HarryMaher: Ideally we wanted two classes - one for donor and another for all donors. This isn't quite the answer. I encourage you to check out what Hosung put in his directory to get an idea of what that would look like

PR #102, opened 2018-05-14
WesWang Mailroom unit test + Circle exercise

WesWang93:
Lesson 6 & Lesson 8

### DIFF
--- a/students/WesWang/lesson8/circle.py
+++ b/students/WesWang/lesson8/circle.py
@@ -1,0 +1,51 @@
+import math
+
+class Circle:
+    def __init__(self, radius):
+        self.radius = radius
+    def __str__(self):
+        return "Circle with radius: " + str(self._radius)
+    def __repr__(self):
+        return f"Circle({self._radius})"
+    def __add__(self, other):
+        total_radius = self._radius + other._radius
+        return Circle(total_radius)
+    def __mul__(self, other):
+        if type(other) == Circle:
+            mul_radius = self._radius * other._radius
+        else:
+            mul_radius = self._radius * other
+        return Circle(mul_radius)
+    def __rmul__(self, value):
+        mul_radius = self._radius * value
+        return Circle(mul_radius)
+    def __eq__(self, other):
+        return self._radius == other._radius
+    def __ne__(self, other):
+        return not self._radius == other._radius
+    def __lt__(self, other):
+        return self._radius < other._radius
+    def __gt__(self, other):
+        return not self._radius < other._radius
+    def __le__(self, other):
+        return self._radius >= other._radius
+    def __ge__(self, other):
+        return not self._radius >= other._radius
+    @property
+    def radius(self):
+        return self._radius
+    @radius.setter
+    def radius(self, radius):
+        self._radius = radius
+    @property
+    def diameter(self):
+        return self._radius*2
+    @diameter.setter
+    def diameter(self, radius):
+        self._radius /= 2
+    @property
+    def area(self):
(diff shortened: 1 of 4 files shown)


students/WesWang/lesson8/circle.py
```diff
@@ -0,0 +1,51 @@
+import math
+
+class Circle:
+    def __init__(self, radius):
+        self.radius = radius
+    def __str__(self):
+        return "Circle with radius: " + str(self._radius)
+    def __repr__(self):
+        return f"Circle({self._radius})"
```
PythonCHB: in the __init__, you use self.radius, but now self._radius 
you should be consitant.

students/WesWang/lesson8/circle.py
```diff
@@ -0,0 +1,51 @@
...
+        return not self._radius == other._radius
+    def __lt__(self, other):
+        return self._radius < other._radius
+    def __gt__(self, other):
+        return not self._radius < other._radius
+    def __le__(self, other):
+        return self._radius >= other._radius
+    def __ge__(self, other):
+        return not self._radius >= other._radius
+    @property
```
PythonCHB: no reason to have a property that simply gets and sets a attribute. YOu can just use self.radius everywhere, and not bother with the property.

Use a property when there is something to be done in the getter and setter -- like compute the diameter :-)

students/WesWang/lesson8/circle.py
```diff
@@ -0,0 +1,51 @@
...
+    def diameter(self):
+        return self._radius*2
+    @diameter.setter
+    def diameter(self, radius):
+        self._radius /= 2
+    @property
+    def area(self):
+        return self._radius**2 * math.pi
+    @classmethod
```
PythonCHB: all good@

But keep style in mind (PEP 8) -- put an empty newline between methods.

students/WesWang/lesson8/test_circle.py
```diff
@@ -0,0 +1,41 @@
...
+
+def test_attr():
+    rad = c1.radius
+    assert rad == 3
+    dia = c1.diameter
+    assert dia == 6
+    area = int(c1.area)
+    assert area == 28
+    
+    c = Circle.from_diameter(10)
```
PythonCHB: better to make this a separate test.

students/WesWang/lesson8/test_circle.py
```diff
@@ -0,0 +1,41 @@
...
+    assert c.radius == 5
+    assert c.diameter == 10
+
+#Test add, multiplication
+def test_math():
+    add = c1 + c2
+    assert add == Circle(7) 
+    mul = c1 * c2
+    assert mul == Circle(12)
+    rmul = 5 * c1
```
PythonCHB: does c1 * 5 work?

students/WesWang/lesson8/test_circle.py
```diff
@@ -0,0 +1,41 @@
...
+    assert c1 < c2 is True
+    assert c1 <= c2 is True
+    assert c1 > c2 is False
+    assert c1 >= c2 is False
+
+def test_sort():
+    clist = [Circle(5), Circle(12), Circle(7), Circle(9), Circle(1), Circle(4), Circle(2)]
+    assert clist[0] == Circle(1)
+    assert clist[3] == Circle(5)
+    assert clist[-1] == Circle(12)
```
PythonCHB: generally better to break things out into more separate tests -- putting a bunch of asserts works fine if they all pass, but if one fails, the ones following it won't run -- so harder to know how much broke....

PR #103, opened 2018-05-14
Lesson 09: OO Mailroom assignment

meet29in:


students/meet_shah/lesson08/Circle.py
```diff
@@ -0,0 +1,55 @@
...
+
+import math
+
+class Circle:
+
+    def __init__(self, radius=0):
+        self._radius = radius
+
+
+    @property
```
PythonCHB: no need to make a property for radius -- unless you really want it to be read-only.

just use a regular attribute.

try testing re-setting the radius -- it will fail -- was that on purpose?

students/meet_shah/lesson08/Circle.py
```diff
@@ -0,0 +1,55 @@
...
+    def radius(self):
+        return self._radius
+
+
+    @property
+    def diameter(self):
+        return (2 * self._radius)
```
PythonCHB: you won't be able to set the diameter, either....

students/meet_shah/lesson08/Circle.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+    @property
+    def area(self):
+        return (self._radius * self._radius * math.pi)
+
+    @classmethod
+    def from_diameter(cls, diameter=0):
```
PythonCHB: nice!

students/meet_shah/lesson08/circle_test.py
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env python3
+
```
PythonCHB: you need to make formal tests here!

example:
```
def test_diameter():
    c - Circle.from_diameter(10)
    assert c.diameter == 10
    assert c.radius == 5
```

and so on -- for all functionality.

students/meet_shah/lesson09/donor_class.py
```diff
@@ -0,0 +1,93 @@
+#!/usr/bin/env python3
+
+from functools import reduce
+
+class Donor:
+
+    def __init__(self, name, donations=None):
+        self._name = name
+        self._donations = [] if donations is None else [donations]
```
PythonCHB: nice!

students/meet_shah/lesson09/donor_class.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+    @property
```
PythonCHB: if it's this simple -- do you need a property?

students/meet_shah/lesson09/donor_class.py
```diff
@@ -0,0 +1,93 @@
...
+    def num_donations(self):
+        return len(self._donations)
+
+
+    @property
+    def total_donations(self):
+        #return reduce(lambda a, x: a+x, self._donations, 0)
+        return sum(self._donations)
```
PythonCHB: kinda makes you see why Guido said: 

"most uses of reduce are just sum() anyway"

:-)

students/meet_shah/lesson09/donor_class.py
```diff
@@ -0,0 +1,93 @@
...
+    def name(self):
+        return self._name
+
+
+    @property
+    def donations(self):
+        return self._donations
+
+
+    def add_donations(self, donation):
```
PythonCHB: careful with naming -- this only adds a single donation -- better to call it "add_donation"

students/meet_shah/lesson09/donor_class.py
```diff
@@ -0,0 +1,93 @@
...
+    def __init__(self):
+        self._donors = {}
+
+
+    def update_donor(self, donor):
+        """
+        update_donor update donor's donation, if a donor DNE, this method adds new one to the list
+        """
+        for donation in donor.donations:
+            self._donors.setdefault(donor.name.lower(), Donor(donor.name)).add_donations(donation)
```
PythonCHB: a donor object can be initialized with a list of donations.

Can you use that to initialize it here without the loop?

students/meet_shah/lesson09/donor_class.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+    def get_total_from_donor(self, donor_name):
+        return self._donors[donor_name.lower()].total_donations
+
+
+    def get_donor(self, donor_name):
+        return self._donors[donor_name.lower()].information
```
PythonCHB: probably better for this to simply return the donor object, then you can get .information from it...

students/meet_shah/lesson09/donor_class.py
```diff
@@ -0,0 +1,93 @@
...
+
+    @property
+    def list_donors(self):
+        return [donor.name for _ , donor in self._donors.items()]
```
PythonCHB: if you aren't using the key, you can do:

`donor.name for donor in self._donors.values()`

students/meet_shah/lesson09/donor_class.py
```diff
@@ -0,0 +1,93 @@
...
+
+    @property
+    def num_donors(self):
+        return len(self._donors)
+
+
+    def generate_sorted_report(self):
+        return sorted([donor.information for _, donor in self._donors.items()], key=lambda item: item[1], reverse=True)
```
PythonCHB: very slick!

students/meet_shah/lesson09/donor_class.py
```diff
@@ -0,0 +1,93 @@
...
+
+    def generate_printable_report(self):
+        return ['{:20}{:>20}{:>12}{:>14}'.format(*donor_info) for donor_info in self.generate_sorted_report()]
+
+
+    def __str__(self):
+        return str(self._donors.keys())
+
+
+    def __repr__(self):
```
PythonCHB: this may be a complex enough object that repr() isn't expected to show everything...

students/meet_shah/lesson09/donor_test.py
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env python3
+
+from donor_class import Donor
+from donor_class import DonorDB
```
PythonCHB: how about:

`from donor_class import Donor, DonorDB`

students/meet_shah/lesson09/donor_test.py
```diff
@@ -0,0 +1,33 @@
...
+def test_donor_init():
+
+    d = Donor('William Gates III')
+    assert d.name == 'William Gates III'
+
+    d.add_donations(10000)
+    d.add_donations(20000)
+
+    assert d.num_donations == 2
+    assert d.total_donations == 30000
```
PythonCHB: nice!

students/meet_shah/lesson09/donor_test.py
```diff
@@ -0,0 +1,33 @@
...
+    #assert d.total_donations() == 30000
+    #assert d.num_donations()
+
+
+def test_donor_db():
+
+    db=DonorDB()
+    mydonor = Donor('William Gates')
+
```
PythonCHB: good start -- but more tests!

can you find a donor in the DB? does it do the right thing if it's not there? if the capitalization is wrong? -- so much more to test for!

students/meet_shah/lesson09/mailroom_oo.py
```diff
@@ -0,0 +1,134 @@
+#!/usr/bin/env python3
```
PythonCHB: really nice job separating out the interaction code from the logic code!

PR #108, opened 2018-05-14
WesWang mailroom fp+oo submission

WesWang93:


students/WesWang/lesson10/mailroom_fp.py
```diff
@@ -0,0 +1,164 @@
...
+
+        if response not in menu:
+            print("Not a Valid Response --- Type 1, 2, 3, or 4")
+        else:
+            menu[response][1]()
+
+
+if __name__ == "__main__":
+    default_donors()
+    mainloop()
```
hosungs: I don't see any functional code (map(), filter(), reduce(), lambda, ...) here, and the required features (projections) are not found, so this submission can't be awarded a complete grade. You can make another submission after implementing the required features with functional code.

students/WesWang/lesson9/mailroom_oo.py
```diff
@@ -0,0 +1,164 @@
...
+
+        if response not in menu:
+            print("Not a Valid Response --- Type 1, 2, 3, or 4")
+        else:
+            menu[response][1]()
+
+
+if __name__ == "__main__":
+    default_donors()
+    mainloop()
```
hosungs: Again, I don't see your code using the defined Donor class at all, so this submission can't be awarded a complete grade. Please do make another submission with expected implementation of your main loop using the defined classes (Donor, DonorDB).

PR #110, opened 2018-05-15
Finished Object Oriented Mailroom Exercise.

Briadean:
Please review OOMailroom.py
Everything else is not finished

Will try to add further code for the circle or functional mailroom exercises shortly.

### DIFF
--- a/students/Briadean/lesson09/OOMailroom.py
+++ b/students/Briadean/lesson09/OOMailroom.py
@@ -1,0 +1,177 @@
+import sys
+
+donors_dict = {"Retsuko Rarecho": [100.00, 25.00, 75.00, 200.00],
+               "Fenneko Inoue": [10.00, 200.00, 75.00],
+               "Haida Kato": [15.00, 15.00, 15.00, 15.00, 15.00],
+               "Gori Tsuruta": [25.00, 30.00, 60.00, 90.00],
+               "Washimi Koiwasaki": [10.00, 10.00, 10.00, 35.00, 15.00]}
+
+
+class Donor:
+    def __init__(self, name, donations=None):
+        self.name = name.title()
+        self.donations = [] if donations is None else donations
+
+    def add_donation(self, donation):
+        self.donations.append(donation)
+
+
+class DonorDB:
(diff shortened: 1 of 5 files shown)


students/Briadean/lesson09/OOMailroom.py
```diff
@@ -0,0 +1,177 @@
...
+        print("Donor List:")
+        for donor in self._donors:
+            print(donor)
+
+    def generate_report(self):
+        """Build a report of donors"""
+        report_lines = []
+        for donor in self._donors.values():
+            total_given = sum(donor.donations)
+            num_gifts = len(donor.donations)
```
hosungs: It would be better to define methods for these total_given and num_gifts in your Donor class...

students/Briadean/lesson09/OOMailroom.py
```diff
@@ -0,0 +1,177 @@
...
+                                                                 "Total Given",
+                                                                 "Num Gifts",
+                                                                 "Average Gifts"
+                                                                 )
+
+        report_lines.sort(key=self._sort_key, reverse=True)
+        report = []
+        report.append(report_header)
+        report.append("-" * 55)
+        [report.append("{:20}  ${:12,.2f}  {:11}  ${:12,.2f}".format(*line)) for line in report_lines]
```
hosungs: Quite an unusual list comprehension usage. I'd just use a simple for loop:

```
for line in report_lines:
    report.append(...)
```

students/Briadean/lesson09/OOMailroom.py
```diff
@@ -0,0 +1,177 @@
...
+        db.add_donor(donor)
+
+
+def main():
+    db = DonorDB()
+    seed_db(db)
+
+    menu_dict = {1: thank_you_menu,
+                 2: db.write_letters_disk,
+                 3: db.generate_report,
```
hosungs: Interesting usage of methods as function names, I'd say, as I myself have never seen this before, but it looks making sense. Maybe this is going to create a curried function, not requiring the self param anyway. Learned something new from this. Thanks.

PR #113, opened 2018-05-15
commit circle.py lesson08

tammytdo:


students/tammyd/lesson08/circle.py
```diff
@@ -0,0 +1,71 @@
...
+        self._radius = value
+
+#Step 4: Add an area property so the user can get the area of the circle
+    @property
+    def area(self):
+        #A = pi*r^2
+        return math.pi * (self._radius ** 2)
+
+#Step 5: Add an “alternate constructor” that lets the user create a Circle directly with the diameter
+    def from_diameter(diameter):
```
hosungs: This needs the `@classmethod` decoration and needs to accept as its first parameter the `cls` parameter. This will not work as is.

PR #114, opened 2018-05-15
Mailroom 5 Updates

kristianjf:
Thank you for all the feedback! I learned a lot through refactoring based on your comments. I was able to achieve the following:

- Removed list/dict data_structure and used objects completely
- Separated user menu (functions) and menu actions (methods)
- Fixed donor initialization (list or int accepted)
- Created separate function for populating DB (test/main use cases)

Remaining Action Items:
- Move user input and logic to functions (normalized_name, thank_you (process name))
- Create unit test based on ability to pass kwargs to menu now

### DIFF
--- a/students/kristianjf/lesson09/mailroom_5.py
+++ b/students/kristianjf/lesson09/mailroom_5.py
@@ -12,44 +12,68 @@ It should have a data structure that holds a list of your donors and a history o
 the amounts they have donated. This structure should be populated at first with at \
 least five donors, with between 1 and 3 donations each.
 '''
-# self.DATA_STRUCTURE = [('Jeff Bezos', [200, 500, 1000]), \
-#                     ('Bill Gates', [1000]), \
-#                     ('Steve Jobs', [20, 50, 100])]
-# self.DATA_STRUCTURE_DICTS = [{'name': name, 'donation': donation} for name, donation in self.DATA_STRUCTURE]
-# # self.DATA_STRUCTURE_DICT = [ \
-# #                         {'name': 'Jeff Bezos', 'donation': [200, 500, 1000]},\
-# #                         {'name': 'Bill Gates', 'donation': [1000]},\
-# #                         {'name': 'Steve Jobs', 'donation': [20, 50, 100]}\
-# #                         ]
-
 class Donor:
     def __init__(self, name, donations):
         self._name = name
-        self._donations = [] if donations is None else donations
+        self._normalized_name = ''
+        self._donations = donations if isinstance(donations, list) else [donations]
+    def __lt__(self, other):
+        return sum(self.donations) < sum(other.donations)
     @property
     def name(self):
         return self._name
     @property
     def donations(self):
         return self._donations
+    @property
+    def normalized_name(self):
+        'Validate and Return Name Input'
+        full_name_cap = ''
+        try:
+            for name in self.name.split():
+                assert name.isalpha()
+                full_name_cap += f'{name.capitalize()} '
+            full_name = full_name_cap.strip()
+        except AssertionError:
+            print('Invalid Name. Found non-alphabetic characters')
+        else:
+            return full_name
+        return self._normalized_name
     def donate(self, new_donation):
         self.donations.append(new_donation)
     def metrics(self):
-        self.num_d = len(self.donations)
-        self.total_d = sum(self.donations)
-        self.avg_d = self.total_d / self.num_d
+        num_d = len(self.donations)
+        total_d = sum(self.donations)
+        avg_d = total_d / num_d
+        return num_d, total_d, avg_d
 
-class DonorsHandle:
-    DATA_STRUCTURE = [] #[('',[])]
-    DATA_STRUCTURE_DICTS = [] #[{'name': None, 'donation': None}]
+    def send_letter(self):
+        '''
+        Try to use a dict and the .format() method to do the letter as one big \
+        template rather than building up a big string in parts.
+        In this version, add a function (and a menu item to invoke it), that goes \
+        through all the donors in your donor data structure, generates a thank you \
+        letter, and writes it to disk as a text file.
+        '''
+        filename = '_'.join(self.name.split())+'.txt'
+        with open(filename, 'w') as outfile:
+            outfile.write(self.output_string().format(name=self.name, donation=self.donations))
+            print(f'Writing: {filename}')
+
+    def output_string(self):
+        'Output string for letters'
+        format_string = 'Dear {name},\n\nThank you for your generous donation of ' \
+                        '${donation}. Please send us more money at your earliest ' \
+                        'convenience.'
+        return format_string
+
+class DonorHandler:
     def __init__(self):
-        pass
+        self.donors = {}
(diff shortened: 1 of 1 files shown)


students/kristianjf/lesson09/mailroom_5.py
```diff
@@ -12,44 +12,68 @@
...
+        return sum(self.donations) < sum(other.donations)
     @property
     def name(self):
         return self._name
     @property
     def donations(self):
         return self._donations
+    @property
+    def normalized_name(self):
+        'Validate and Return Name Input'
```
hosungs: Use proper docstring commenting delimiters (`'''...'''`)

students/kristianjf/lesson09/mailroom_5.py
```diff
@@ -12,44 +12,68 @@
...
+        through all the donors in your donor data structure, generates a thank you \
+        letter, and writes it to disk as a text file.
+        '''
+        filename = '_'.join(self.name.split())+'.txt'
+        with open(filename, 'w') as outfile:
+            outfile.write(self.output_string().format(name=self.name, donation=self.donations))
+            print(f'Writing: {filename}')
+
+    def output_string(self):
+        'Output string for letters'
```
hosungs: Docstring

students/kristianjf/lesson09/mailroom_5.py
```diff
@@ -12,44 +12,68 @@
...
+        format_string = 'Dear {name},\n\nThank you for your generous donation of ' \
+                        '${donation}. Please send us more money at your earliest ' \
+                        'convenience.'
+        return format_string
```
hosungs: This return value looks like a fixed string, so I'm not sure if this needs to be a method? Could have been just a class attribute?

students/kristianjf/lesson09/mailroom_5.py
```diff
@@ -201,21 +162,75 @@ def name_response_valid(self, full_name):
...
+
+#Thank You Menu
+def thank_you_menu(**kwargs):
+    '''
+    Sending a Thank You
+    If the user (you) selects ‘Send a Thank You’, prompt for a Full Name.
+    If the user types ‘list’, show them a list of the donor names and re-prompt
+    '''
+    program_options_ty = {'Send a Thank You': kwargs['donor_handler'].thank_you, \
+                        'Action: List': kwargs['donor_handler'].list_donors, \
```
hosungs: I suppose you could have just passed the donor_handler object as a usual argument, not as kwargs, but I may be nitpicking anyway... :) Just wondering if there's any special reason for you to choose kwargs rather than a normal argument?

students/kristianjf/lesson09/mailroom_5.py
PythonCHB: good note -- technically, a docstring is just a string, so single quotes work, but it's considered good style to use triple quotes even if it's a one-liner docstring -- that way it's easier to extend.

students/kristianjf/lesson09/mailroom_5.py
```diff
@@ -64,38 +88,39 @@ def thank_you(self, full_name='', donation_amount=int()):
...
+    def list_donors(self, **kwargs):
         '''
         If the user types ‘list’, show them a list of the donor names and re-prompt
         '''
-        for entry in self.DATA_STRUCTURE:
-            print(entry[0])
-        return [entry[0] for entry in self.DATA_STRUCTURE]
+        for name in self.donors:
+            print(name)
+        return [name for name in self.donors]
```
PythonCHB: I think I've seen this before -- DRY! (even though it's only one line)

students/kristianjf/lesson09/mailroom_5.py
```diff
@@ -64,38 +88,39 @@ def thank_you(self, full_name='', donation_amount=int()):
...
 
-    def list_donors(self):
+    def list_donors(self, **kwargs):
         '''
         If the user types ‘list’, show them a list of the donor names and re-prompt
         '''
-        for entry in self.DATA_STRUCTURE:
-            print(entry[0])
-        return [entry[0] for entry in self.DATA_STRUCTURE]
+        for name in self.donors:
```
PythonCHB: don't loop through to print, then again to return a list....

and if you are returning the list -- why print it here?

students/kristianjf/lesson09/mailroom_5.py
```diff
@@ -106,87 +131,23 @@ def create_a_report(self):
...
-        except ValueError:
-            print('Non-integer value entered. Please try again.')
-        except AssertionError:
-            print('This number does not exist in the list of options. Please try again.')
-        else:
-            return response
+        for donor in self.donors.values():
+            donor.send_letter()
     #Name Validation
     def name_response_valid(self, full_name):
```
PythonCHB: haven't I seen this before? -- DRY!!!

students/kristianjf/lesson09/mailroom_5.py
PythonCHB: agreed -- better to pass the dict in as a regular parameter.

students/kristianjf/lesson09/mailroom_5.py
```diff
@@ -201,21 +162,75 @@ def name_response_valid(self, full_name):
...
+    Sending a Thank You
+    If the user (you) selects ‘Send a Thank You’, prompt for a Full Name.
+    If the user types ‘list’, show them a list of the donor names and re-prompt
+    '''
+    program_options_ty = {'Send a Thank You': kwargs['donor_handler'].thank_you, \
+                        'Action: List': kwargs['donor_handler'].list_donors, \
+                        'quit': main_menu}
+    menu(program_options_ty, **kwargs)
+
+def quit_menu(**kwargs):
```
PythonCHB: what might you be passing in here? that is, why the **kwargs?

students/kristianjf/lesson09/mailroom_5.py
```diff
@@ -201,21 +162,75 @@ def name_response_valid(self, full_name):
...
+    sys.exit()
+
+#Menu Option Validation
+def menu_response_valid(options):
+    'Validate Menu Options'
+    try:
+        response = int(input())
+        assert response in [option[0] for option in options]
+    except ValueError:
+        print('Non-integer value entered. Please try again.')
```
PythonCHB: is it necessarily an integer at this point?

PythonCHB:
Enjoy the next class!

students/kristianjf/lesson09/mailroom_5.py
kristianjf: Thanks for the feedback. In retrospect, passing kwargs was not necessary. I believe I previously attempted to pass kwargs outside the class. Now looking at the code, kwargs for the menu may be an unnecessary complication.

Thanks!

students/kristianjf/lesson09/mailroom_5.py
PythonCHB: ah, then you want to put the except right after the `response = int(input())` line -- you always want the except as close as possible to where the error may get raised.

PR #115, opened 2018-05-15
commit mailroom_oo.py

tammytdo:


students/tammyd/lesson09/mailroom_oo.py
```diff
@@ -0,0 +1,235 @@
...
+    #pt1: a history of the amounts they have donated.
+    @property
+    def total_donations(self):
+        return sum(self._donations)
+
+    @property
+    def new_donation(self, donation):
+        self._donations.append(donation)
+
+    @property
```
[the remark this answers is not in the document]
PythonCHB: actually, this on is OK -- total_donations as a property makes sense, then you can do:

print("the total donated by this donor is:", donor.total_donations)

-CHB

students/tammyd/lesson09/mailroom_oo.py
```diff
@@ -0,0 +1,235 @@
...
+    def avg_donations(self):
+        return sum(self._donations) / len(self._donations)
+
+
+
+
+#You’ll then want a class that handles the collection of donors. 
+#method needs the information from the whole collection.
+#pt1: data structure that holds a list of your donors
+#Do I want to include or exclude () right after DonorDB here?
```
PythonCHB: no need for parens here -- you use parens when you subclass from something.

students/tammyd/lesson09/mailroom_oo.py
```diff
@@ -0,0 +1,235 @@
...
+class DonorDB:
```
PythonCHB: fine start -- but you need a lot more here!

students/tammyd/lesson09/mailroom_oo.py
```diff
@@ -0,0 +1,235 @@
...
+        # else:
+        #     break
+        except KeyError:
+            print("Incorrect entry. Try again.")
+            return
+    donation(donor_name)
+
+
+
+def find_donor(donor_name):
```
PythonCHB: this should be a method of your DonorDB class

students/tammyd/lesson09/mailroom_oo.py
```diff
@@ -0,0 +1,235 @@
...
+    for donor in donor_db:
+        if donor_name.strip().lower() == donor[0].lower():
+            return donor
+    return None
+
+def ty_letter(donor_name, donor_amount):
```
PythonCHB: this should be a method of the Donor class.

students/tammyd/lesson09/mailroom_oo.py
```diff
@@ -0,0 +1,235 @@
...
+    print("\nMenu: Send a Thank You, Letter\n")
+    letter = (f"\nDear {donor_name.title()},\n\n\tYour donation of ${donor_amount} will make a great difference to a young learner at the coding academy. Thank you.\n\n\nSincerely,\nAn Academy Student\n")
+    print(letter)
+    print("--------------------------------\n\n")
+    main_menu()
+
+
+def letters_to_everyone():
+    for donor_name in donor_db.values():
+        letter = (f"\nDear {donor_name[0]},\n\n\tYour donation of ${donor_name[1][-1]} will make a great difference to a young learner at the coding academy. Thank you.\n\n\nSincerely,\nAn Academy Student\n")
```
PythonCHB: this is the same code as the previous function! -- you should call that from this function.

students/tammyd/lesson09/mailroom_oo.py
```diff
@@ -0,0 +1,235 @@
...
+
+def ty_letter(donor_name, donor_amount):
+    print("\nMenu: Send a Thank You, Letter\n")
+    letter = (f"\nDear {donor_name.title()},\n\n\tYour donation of ${donor_amount} will make a great difference to a young learner at the coding academy. Thank you.\n\n\nSincerely,\nAn Academy Student\n")
+    print(letter)
+    print("--------------------------------\n\n")
+    main_menu()
+
+
+def letters_to_everyone():
```
PythonCHB: method of DonorDB class...

are you getting the idea?

tammytdo:
Hello Chris,

Thank you immensely. I am committed to re-studying all the material and
comments, and getting extra support before the Advanced course begins.

With gratitude,
Tammy

PythonCHB:
I"m merging this, and giving you credit for OO mailroom, to get yu over the 80% ,mark, so you can move on to the next class.

But you really need to get basic OO down at some point -- it's only going to get harder next quarter, so I encourage you to work through both the HTML render and OO mailroom exercises -- if yo8u can get those cleanly done, you'll be in much better shape for moving through the rest of the program.

students/tammyd/lesson09/mailroom_oo.py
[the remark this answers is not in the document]
hosungs: Oh, yes. Sorry for misleading. I must have gone through this PR too fast in my morning commute bus... Yes, this is OK, and actually this is how I did it in class...

PR #116, opened 2018-05-16
 fp: mailroom assignment

kristianjf:


PythonCHB:
This looks , at a glance like the same code. 

Anything new you want reviewed?

kristianjf:
Added the following methods:
- dnr_challenge
- dnr_filter
- dnr_projections
- run_projection

Updated test module with the following functions:
- test_donor_challenge
- test_donor_projection

I spent quite a bit time on the run_projection method attempting to create a dictionary switch that would allow me to list the donors as the key and updating the kwargs as the value. I struggled to get past the first step in the thought process below:

> In run_projection method:
> 1. Create a dictionary switch that gets passed to the menu function that has list of donors as key and the value updates kwargs dictionary with the name of the selected donor
> 2. Once kwargs has the selected donor, collect inputs for the max_donation_amount and factor.
> 3. Call the dnr_projections method.

```python
'Select from list of Donors to Run Projections'
    def run_projection(self, full_name='', max_donation_amount=int(), factor=int(), **kwargs):
        'Select from list of Donors to Run Projections'
        donor_list_dict = {donor: kwargs.update({'full_name': donor}) for donor in self.donors}
        menu(donor_list_dict)
```

I received the follow error:
`TypeError: 'NoneType' object is not callable`

Any advice you can offer on how to work past this would be appreciated.

Lastly, quick question regarding the map function. I thought it was odd that I had to map a list (x) to a list of the same length with the values equal to the multiplier (y). Is this common?

```python
            map_donations_f = list(map(lambda x, y: x*y, filter_donations,\
            len(filter_donations)*[factor]))
```

kristianjf:
@PythonCHB , any comments?

PythonCHB:
I kinda forgot about this once it got merged -- but i'll take a look now.

PythonCHB:
for this:

`map_donations_f = list(map(lambda x, y: x*y, filter_donations, len(filter_donations)*[factor]))`
you should be able to pass factor in without making another list out of it.

```
In [2]: factor = 5

In [3]: donations = [5, 6, 7]

In [4]: list(map(lambda x: factor*x, donations))
Out[4]: [25, 30, 35]
```
that is, the name "factor" is in that namespace.

so try:

`map_donations_f = list(map(lambda x: x * factor, filter_donations))`

students/kristianjf/lesson10/mailroom_6.py
```diff
@@ -0,0 +1,302 @@
...
+'''
+It should have a data structure that holds a list of your donors and a history of \
+the amounts they have donated. This structure should be populated at first with at \
+least five donors, with between 1 and 3 donations each.
+'''
+class Donor:
+    def __init__(self, name, donations):
+        self._name = name
+        self._normalized_name = ''
+        self._donations = donations if isinstance(donations, list) else [donations]
```
PythonCHB: you want be a bit careful here -- maybe be able to accept other sequences than lists? or even other iterables?

you could do:

```
try:
    self._donations = list(donations)
except TypeError:
    self._donations = [donations]
```

students/kristianjf/lesson10/mailroom_6.py
```diff
@@ -0,0 +1,302 @@
...
+    def __lt__(self, other):
+        return sum(self.donations) < sum(other.donations)
+    @property
+    def name(self):
+        return self._name
+    @property
```
PythonCHB: why not just have these regular attributes?

students/kristianjf/lesson10/mailroom_6.py
```diff
@@ -0,0 +1,302 @@
...
+    def donations(self):
+        return self._donations
+    @property
+    def normalized_name(self):
+        'Validate and Return Name Input'
+        full_name_cap = ''
+        try:
+            for name in self.name.split():
+                assert name.isalpha()
```
PythonCHB: you don't want an assertion here -- asserts are for testing.

somethign like:

```
if not name.isapha():
    raise ValueError(names can only have alphabetical characters")
```

but this is not where you should test validity anyway :-)

students/kristianjf/lesson10/mailroom_6.py
```diff
@@ -0,0 +1,302 @@
...
+        for option in options:
+            print(option)
+        response = menu_response_valid(options)
+        response_selection = options[response-1][1]
+        options_dict[response_selection](**kwargs)
+
+#Main Menu
+def main_menu(**kwargs):
+    'Create Main Menu'
+    program_options_dict = {'Send a Thank You': thank_you_menu, \
```
PythonCHB: no need for slashes here -- they aren't required if there an unclosed brackets.

students/kristianjf/lesson10/mailroom_6.py
```diff
@@ -0,0 +1,302 @@
...
+    Sending a Thank You
+    If the user (you) selects ‘Send a Thank You’, prompt for a Full Name.
+    If the user types ‘list’, show them a list of the donor names and re-prompt
+    '''
+    program_options_ty = {'Send a Thank You': kwargs['donor_handler'].thank_you, \
+                        'Action: List': kwargs['donor_handler'].list_donors, \
+                        'quit': main_menu}
+    menu(program_options_ty, **kwargs)
+
+def quit_menu(**kwargs):
```
PythonCHB: really not sure why you are passing kwargs around everywhere???

students/kristianjf/lesson10/mailroom_6.py
```diff
@@ -0,0 +1,302 @@
...
+            len(filter_donations)*[factor]))
+
+            new_d = Donor(donor, map_donations_f)
+            new_dh.add_donor(new_d)
+        return new_dh
+    def dnr_filter(self, lst, min_donation=0, max_donation=10000000):
+        lst = list(filter(lambda x: (x > min_donation) and (x < max_donation), lst))
+        return lst
+
+    def dnr_projections(self, donor, **kwargs):
```
PythonCHB: I don't think kwargs is the way to pass info around like this.

if it's persistent info -- store it in an attribute, if it's ephemeral, pass it in a dict or other data structure.

students/kristianjf/lesson10/mailroom_6.py
```diff
@@ -0,0 +1,302 @@
...
+            donations of all the donors by the factor. The function returns a NEW donor database,
+            with the new data.'''
+            map_donations_f = list(map(lambda x, y: x*y, filter_donations,\
+            len(filter_donations)*[factor]))
+
+            new_d = Donor(donor, map_donations_f)
+            new_dh.add_donor(new_d)
+        return new_dh
+    def dnr_filter(self, lst, min_donation=0, max_donation=10000000):
+        lst = list(filter(lambda x: (x > min_donation) and (x < max_donation), lst))
```
PythonCHB: do you need to make a list? if it's going to be looped over, it could stay an iterable.
that's why map, filter, zip, etc return iterables -- so they can be looped over without creating a whole new sequence.

students/kristianjf/lesson10/mailroom_6.py
```diff
@@ -0,0 +1,302 @@
...
+class DonorHandler:
+    def __init__(self):
+        self.donors = {}
+    def add_donor(self, donor_object):
+        self.donors[donor_object.normalized_name] = donor_object
+    def dnr_challenge(self, factor=1, **kwargs):
+        new_dh = DonorHandler()
+        for donor, donor_object in self.donors.items():
+            # new_d = Donor(donor, list(map(lambda x, y: x*y, \
+            # donor_object.donations, len(donor_object.donations)*[factor])))
```
PythonCHB: Maybe this should be a method of the Donor object? so you can make a copy, and then apply a factor to it?

students/kristianjf/lesson10/mailroom_6.py
```diff
@@ -0,0 +1,302 @@
...
+            # new_d = Donor(donor, \
+            # self.dnr_filter(\
+            # list(map(lambda x, y: x*y, donor_object.donations, \
+            # len(donor_object.donations)*[factor])),\
+            # kwargs['min_donation'], kwargs['max_donation']))
```
PythonCHB: rather than kwargs, just pass in min_donation and max_donation -- or a tuple maybe?

students/kristianjf/lesson10/mailroom_6.py
```diff
@@ -0,0 +1,302 @@
...
+            # new_dh.add_donor(new_d)
+            '''Add a new feature to Mailroom using filter so that donations either above
+            or below a specified dollar amount are included in the map operations of #1 above.
+            You can do this by adding min_donation and max_donation optional keyword parameters
```
PythonCHB: this means something like:

`def challenge( min_donation=None, max_donation=None):`

that's more clear what you need than simply expecting it to be in kwargs....

PythonCHB (commented):
could use a little cleanup, but you've got the ideas of map and filter down.

PythonCHB:
Enjoy the next class!

-CHB

kristianjf:
Thank you!
Will try to keep these lessons in mind as I proceed:

- Use assertion for testing, try and except for catching and normalizing inputs.
- Keep in mind ways to allow flexibility of different sequence types
- Where possible, be explicit (by setting keyword arguments in the function names) rather than implicit (passing kwargs dictionary).

Thanks for the feedback!!

PR #117, opened 2018-05-17
Long-overdue pull request 

aravichander:
Playing catch-up


students/aravichander/lesson08/circle_class.py
```diff
@@ -0,0 +1,54 @@
...
+	@property
+	def area(self):
+		return pi*(self.radius**2)
+		#print("Area of circle is {0:.2f}".format(area))
+
+	#not fully understanding this but hoping I did this correctly!
+	@classmethod
+	def circle_from_diameter(cls, diameter):
+		radius = diameter/2
+		return cls.Circle(radius)
```
PythonCHB: not quite, but close -- in a classmethod, the first parameter (cls) is the class object itself -- in this case "Circle", so you can call it to make a new one:

``return cls(radius)``

make sure to write a test for this -- it won't run like it is!

students/aravichander/lesson08/circle_class.py
```diff
@@ -0,0 +1,54 @@
+from math import pi 
+circlelist = []
+
+#Creating class to represent circle object 
+class Circle:
+	def __init__(self,radius):
```
PythonCHB: use four spaces to indent -- always! 

I think these are tabs. Make sure your editor is configured to put in four spaces when you hit the <tab> key.

students/aravichander/lesson08/circle_class.py
```diff
@@ -0,0 +1,54 @@
...
+			self.radius = radius
+			print("Circle with radius {0:.6f}".format(radius))
+	
+	@property
+	def diameter(self):
+		return self.radius*2
+
+	@diameter.setter
+	def diameter(self,diameter):
+		self.radius = (diameter)/2
```
PythonCHB: no need for parens here.

students/aravichander/lesson08/circle_class.py
```diff
@@ -0,0 +1,54 @@
...
+	#not fully understanding this but hoping I did this correctly!
+	@classmethod
+	def circle_from_diameter(cls, diameter):
+		radius = diameter/2
+		return cls.Circle(radius)
+	
+	#being able to add 2 new circles together
+	def __add__(self,newcircle):
+		if isinstance(newcircle,Circle):
+			return self.radius + newcircle.radius
```
PythonCHB: this is going to return a number -- the sum of the radii. but you want it to return a circle object:

``Circle(self.radius + newcircle.radius)``

But you don't want to check to see if it's an instance of Circle -- just use the radius. You'll get a TypeError if "newcircle" doesn't have a radius -- which is what you want.

students/aravichander/lesson08/test_circle_class.py
```diff
@@ -0,0 +1,6 @@
+def test_circle_size():
```
PythonCHB: good start, but more tests! 

what about setting the diameter? and creating one from a diameter? and adding them together, and....

PythonCHB: Remember test driven development -- you should have a test for the feature you are working on as soon as you start to write the code....

students/aravichander/lesson03/strformat_lab.py
```diff
@@ -0,0 +1,35 @@
...
+# 	listlen = len(numbers)
+# 	return "The",listlen,"numbers are {0}".format(numbers)
+
+# print(multipletuple(1,2))
+
+#Task Four
+tuple3 = (4, 30, 2017, 2, 27)
+#print("{0[3]},{0[4]},{0[2]},{0[0]},{0[1]}".format(tuple3))
+
+#Task Five - don't have Python 3.6
```
PythonCHB: do install 3.6 -- it's got some nice features -- like this one :-)

students/aravichander/lesson03/list_lab.py
```diff
@@ -0,0 +1,59 @@
...
+# Fruitdelete = input("Which fruit do you want to remove?\n")
+# DelPosition = AllFruits.index(Fruitdelete)
+# del AllFruits[DelPosition]
+# print(AllFruits)
+
+#Series3
+
+AllFruits2 = AllFruits
+for i in range(0,listlen):
+	response4 = input("Do you like" + AllFruits[i] + "?")
```
PythonCHB: 4 spaces for indent!

PR #119, opened 2018-05-21
Modifications to lesson06

aravichander:


students/aravichander/lesson06/Mailroom3.py
```diff
@@ -0,0 +1,89 @@
...
+
+Donors = {"Bill Gates":[10000,12000],
+		  "Jeff Bezos":[50],`
+		  "Mark Zuckerberg": [500,600,700],
+		  "Paul Allen": [250,350,450],
+		  "King of Siam": [200,250]
+		  }
+
+def printdonors():
+	for x,v in Donors.items():
```
PythonCHB: looks like you have <tab> characters here.

you REALLY want to use four spaces. Make sure your editor is set to insert spaces when you hit the <tab> key!

students/aravichander/lesson06/Mailroom3.py
```diff
@@ -0,0 +1,89 @@
...
+		print(x,"donated $",v)
+
+def thankall():
+	for x in Donors:
```
PythonCHB: you are working with botht he key and the value, so better to use:

```
for donor, donations in Donors.items():
    letterofthanks(donor, sum(donations))
```

and avoid using 'x' when there is a more meaningful name available.

students/aravichander/lesson06/Mailroom3.py
```diff
@@ -0,0 +1,89 @@
...
+	print("\nLetter of thanks has been saved to hard disk as text file under "+str(filename))
+
+def searchdonors(thankdonor):
+	#Not sure if these exceptions here are needed to be honest#
+	for x in Donors:
+		try:
+			if thankdonor == x:
+				totaldonated = sum(Donors[x])
+				letterofthanks(thankdonor,totaldonated)
+		except KeyError:
```
PythonCHB: you are right -- not helpful here -- the key is gotten from the for loop, so it is in the dict, and you won't ever get a KeyError.

students/aravichander/lesson06/Mailroom3.py
```diff
@@ -0,0 +1,89 @@
...
+			print("response was,",response)
+			if response =="1":
+				print("\nYou have chosen the 'Thank' option")
+				thanks()
+			elif response == "2":
+				report()
+			elif response == "3":
+				thankall()
+			else:
+				print("\nSee you later!")
```
PythonCHB: this is going to exit if they make a typo -- it should probably return to the prompt in that case, and only exit if they specifically ask to quit (typing 4, or "quit", or ...)

students/aravichander/lesson07/Mailroom4.py
```diff
@@ -0,0 +1,88 @@
+#Mailroom #2
```
PythonCHB: what's new here?

be sure to let us know what you've done, so we can review it.

PythonCHB:
I'm going to merge, as this has all been reviewed.

You can make a new PR when you have something new to submit.

PR #120, opened 2018-05-23
Added exceptions for Mailroom 3 + unit-testing and refactored functions for Mailroom 4 

aravichander:


students/aravichander/lesson06/Mailroom3.py
```diff
@@ -1,19 +1,21 @@
...
 import unittest 
 
-Donors = {"Bill Gates":[10000,12000],
-		  "Jeff Bezos":[50],`
-		  "Mark Zuckerberg": [500,600,700],
-		  "Paul Allen": [250,350,450],
-		  "King of Siam": [200,250]
-		  }
+Donors = {}
+def createlist():
```
PythonCHB: better to create the dict in the function and return it, rather than having it as a global.

PythonCHB: use names that describe what the function does:

this is crating a dict, not a list :-) -- but also -- what kind of list(dict?) -- somethign like:

`create_sample_donors()`

would be more clear.

students/aravichander/lesson06/Mailroom3.py
```diff
@@ -1,19 +1,21 @@
...
+	Donors["Bill Gates"] = [10000,12000]
+	Donors["Jeff Bezos"]=[50]
+	Donors["Mark Zuckerberg"] = [500,600,700]
+	Donors["Paul Allen"] = [250,350,450]
+	Donors["King of Siam"] = [200,250]
+	
 
 def printdonors():
```
PythonCHB: see PEP 8: underscores help make names easier ot read:

`print_donors`

students/aravichander/lesson06/Mailroom3.py
```diff
@@ -1,19 +1,21 @@
...
 	for x,v in Donors.items():
-		print(x,"donated $",v)
+		print("\n{} donated ${}".format(x,v))
```
PythonCHB: this will all be easeir to test if the functions that create text return a string -- then you can test that (and potentially use it in other ways) -- when you want to print, you can keep that isolated:

`print(donor_listing)` or something like that.

students/aravichander/lesson06/Mailroom3.py
```diff
@@ -30,60 +32,70 @@ def letterofthanks(donorname,totaldonated):
 	print("\nLetter of thanks has been saved to hard disk as text file under "+str(filename))
 
 def searchdonors(thankdonor):
-	#Not sure if these exceptions here are needed to be honest#
-	for x in Donors:
 		try:
-			if thankdonor == x:
-				totaldonated = sum(Donors[x])
-				letterofthanks(thankdonor,totaldonated)
+			print(Donors[thankdonor])
```
PythonCHB: as above -- you don't want your searching function to print anything -- it should return the donor if found, and either None or raise an exception is not found.

students/aravichander/lesson07/Mailroom_unittest.py
```diff
@@ -0,0 +1,33 @@
+import unittest
```
PythonCHB: You want to name your test files something like: `test_mailrroom.py`. that way test runners like pytest will find it.

PythonCHB:
For the OO version -- this is a really good time to use test driven development.

Before you do any more -- create a file called test_mailroom.py, and start putting tests in there.

Right now, you can test creating a Donor, and testing that the name is set correctly. 

Then you can write a couple tests for the DonorList class, ans make sure that the properties you have defined work correctly.

Then you can go through your existing functions, and move them one by one into methods of your classes -- writing a test or two for each one as you do.

If a function works with a single donor, put it in the Donor class

If a function works with multiple donors, put it in the DonorList class.

If a function has an `input()` in it, keep it out of the classes -- or create a CLI class for them (Command Line Interface).

If there is any logic mixed in with the functions with `input` -- time to split that out.

If there are any `print()`s in the logic classes, remove them and add them to the CLI functions.

You can focus on this now -- if you have good tests for OO mailroom, we'll give you credit for tests and OO :-)

students/aravichander/lesson07/Mailroom_unittest.py
```diff
@@ -0,0 +1,33 @@
...
+
+from Mailroom_lesson07 import createlist,getlist,sumdonor
+
+class MyFuncTestCase(unittest.TestCase):
```
PythonCHB: you are welcome to use unittest if you want -- but it's extra overhead for little gain. That's why we teach and recommend using pytest and simple test functions.

for example, these tests would be as simple as:

```
def test_createlist(self):
    # note -- this kinda shows why you want createlist() to return the list...
    createlist()
    output = getlist()
    output["Bill Gates"] == [10000,12000] 

def test_find_not_there():
    with self.assertRaises(ValueError):
        searchdonors("Carrot Top")
# though I'd probably have searchdonors() return None if it's not there, rather than raising an error.

def test_sumdonor(self):
    # note that you are changing a global with each test run -- not good. which is why you want it to return data, rather than altering a global. This all gets easier with OO -- that's one reason to do it :-)
    createlist()
    sumdonor("Paul Allen") == 1050
```

students/aravichander/lesson07/Mailroom_unittest.py
```diff
@@ -0,0 +1,33 @@
...
+	def test_createlist(self):
+		createlist()
+		output = getlist()
+		self.assertEqual(output["Bill Gates"],[10000,12000]) 
+		self.assertTrue(output["Bill Gates"] == [10000,12000]) 
+		self.assertFalse(output["Bill Gates"] == [10000,12001])
+		with self.assertRaises(KeyError):
+			output["Carrot Top"]
+
+#This was a failed testing scenario as the original function already handles the exception
```
PythonCHB: so this is a good hint that you want to refactor your code :-)

you want a function (or method, when you go OO) that finds a given donor -- 

What should it do if it finds the donor? probably return the donor dat structure or object.

What should it do if the donor is not there? It should either return a "null" value -- i.e. None, or raise an exception. Once you decide that, then you know what to test for.

students/aravichander/lesson07/Mailroom_unittest.py
```diff
@@ -0,0 +1,33 @@
...
+from Mailroom_lesson07 import createlist,getlist,sumdonor
+
+class MyFuncTestCase(unittest.TestCase):
+	
+	def test_createlist(self):
+		createlist()
+		output = getlist()
+		self.assertEqual(output["Bill Gates"],[10000,12000]) 
+		self.assertTrue(output["Bill Gates"] == [10000,12000]) 
+		self.assertFalse(output["Bill Gates"] == [10000,12001])
```
PythonCHB: you don't need to test non-equality -- if the result is correct, then it's correct. So the assertEqual is all you need here.

students/aravichander/lesson07/Mailroom_unittest.py
```diff
@@ -0,0 +1,33 @@
...
+
+from Mailroom_lesson07 import createlist,getlist,sumdonor
+
+class MyFuncTestCase(unittest.TestCase):
+	
+	def test_createlist(self):
+		createlist()
+		output = getlist()
+		self.assertEqual(output["Bill Gates"],[10000,12000]) 
+		self.assertTrue(output["Bill Gates"] == [10000,12000])
```
PythonCHB: If you are using `unittest`, you want to use the custom asserts (i.e. `assertEqual`) whenever possible. Doing an `assertTrue` with a `==` is exactly the same, but with less useful output on a failure.

students/aravichander/lesson07/Mailroom_unittest.py
```diff
@@ -0,0 +1,33 @@
...
+#This was a failed testing scenario as the original function already handles the exception
+	# def test_searchdonors(self):
+	# 	createlist()
+	# 	output = getlist()
+	# 	with self.assertRaises(KeyError):
+	# 		searchdonors("Carrot Top")
+
+	def test_sumdonor(self):
+		createlist()
+		self.assertTrue(sumdonor("Paul Allen") == 1050)
```
PythonCHB: again -- assertEqual, if you are using unittest.

and no need to test False  -- test for what you want -- if it gives the right result, it gives the right result.

students/aravichander/lesson07/Mailroom_unittest.py
```diff
@@ -0,0 +1,33 @@
...
+		self.assertFalse(sumdonor("Jeff Bezos") == 250000000)
+		with self.assertRaises(KeyError):
```
PythonCHB: checking for a non-existent donor should be a separate test.

students/aravichander/lesson09/Mailroom_class.py
```diff
@@ -0,0 +1,146 @@
...
+class Donor:
+	def __init__(self,Name):
+		self.Name = Name
+
+	@property
+	def donations(self):
+		return self.donations
+
+	@classmethod
+	def sumdonation(cls, donations):
```
PythonCHB: you don't want a classmethod here -- the sum of donations is a simple number that should be returned. 

a property is good in this case.

classmethods are for creating a new instance of a class for some reason -- no need for that here.

PythonCHB:
I'm going to merge this, as it's been reviewed.

You can do a new PR when you have new code to review.

I've given you credit for mailroom3 -- exceptions -- though do refactor those to be actually useful in your OO version. You won't have a lot :-)

PR #121, opened 2018-05-24
Mailroom_testing added along with object-oriented Mailroom (Mailroom_class)

aravichander:


students/aravichander/lesson08/circle_class.py
```diff
@@ -9,7 +9,7 @@ def __init__(self,radius):
 		elif radius < 0:
 			raise ValueError
```
PythonCHB, line 10: correct Exception to raise, but usually good to add a message:

`raise ValueError("radius must be a positive value")`

students/aravichander/lesson08/circle_class.py
```diff
@@ -20,24 +20,23 @@ def __repr__(self):
...
 		#print("Area of circle is {0:.2f}".format(area))
 
 	#not fully understanding this but hoping I did this correctly!
+
 	@classmethod
-	def circle_from_diameter(cls, diameter):
-		radius = diameter/2
-		return cls.Circle(radius)
+	def from_diameter(cls, diameter):
+		return Circle(diameter / 2.0)
```
PythonCHB, line 38: almost! this will indeed work, but kind of bypasses the class method.  in this context, `cls` IS the class object, so you can do:

`return cls(diameter / 2.0)`

In this case, this will result in exactly the same thing, but if you use `cls`, then this will ALSO work with a subclass of Circle.

students/aravichander/lesson08/circle_class.py
```diff
@@ -20,24 +20,23 @@ def __repr__(self):
...
 
-	
 	#being able to add 2 new circles together
 	def __add__(self,newcircle):
 		if isinstance(newcircle,Circle):
```
PythonCHB, line 42: I'd probably only allow this to work with a Circle object (or other object with a radius).

students/aravichander/lesson09/Mailroom_class.py
```diff
@@ -3,68 +3,105 @@
...
-	@classmethod
-	def sumdonation(cls, donations):
-		total = sum(self.donations)
-		return cls.Donor(Name)
-
-class Donorlist:
-	def __init__(self,Donors)
 
+class Donor:
+	def __init__(self,name,donations = None): #add default parameters for donations
```
PythonCHB, line 10: keep an eye on PEP8 for style:

`def __init__(self, name, donations=None):`

seems trivial, but it does make your code more readable to have consistent spacing.

students/aravichander/lesson09/Mailroom_class.py
```diff
@@ -3,68 +3,105 @@
...
+		self._name = name
+		self._donations = [] if donations is None else donations
```
PythonCHB, line 12: nice!

students/aravichander/lesson09/Mailroom_class.py
```diff
@@ -3,68 +3,105 @@
...
-		letterofthanks(x,totaldonated)
+	@property
+	def sum_donation(self):
+		return sum(self._donations)
 
-def letterofthanks(donorname,totaldonated):
+	@property
+	def number_of_donations(self):
+		return len(self._donations)
+#My ipython was going nuts so I had to manually calculate the average >=(
```
PythonCHB, line 32: huh? the following is fine.

students/aravichander/lesson09/Mailroom_class.py
```diff
@@ -3,68 +3,105 @@
...
+	@property
+	def donation_average(self):
+		return sum(self._donations)/len(self._donations)
```
PythonCHB, line 35: you should be able to use self.number_of_donations here, but this is fine, too.

students/aravichander/lesson09/Mailroom_class.py
```diff
@@ -3,68 +3,105 @@
...
+
+
+class Donor_db:
+	def __init__(self):
+		self._donors = {}
+
+	def add_donor(self,donor):
+		self._donors[donor.name.lower()] = donor
+
+	# def all_donors_donations(self): Couldn't figure out the issue here re:testing
```
PythonCHB, line 46: hmm -- looks OK to me.

though it could be more compact:

`total_sum = sum((d.sum_donation for d in self._donors.items()))`

or something like that -- untested

students/aravichander/lesson09/Mailroom_class.py
```diff
@@ -3,68 +3,105 @@
...
+	# 	total_sum = 0
+	# 	for name,donor_obj in self._donors.items():
+	# 		print(name)
+	# 		total_sum = total_sum + donor_obj.sum_donation
+	# 	return total_sum
+
+	def get_donor(self, donor_name):
```
PythonCHB, line 53: good -- this could be made "fancier" in the future -- so nice to put it in a method.

students/aravichander/lesson09/Mailroom_class.py
```diff
@@ -3,68 +3,105 @@
...
+	 	return self._donors[donor_name.lower()]
+
+	def print_Db(self):
+		for name,donor_obj in self._donors.items():
```
PythonCHB, line 57: look out for those <tab> characters!

students/aravichander/lesson09/Mailroom_class.py
```diff
@@ -3,68 +3,105 @@
...
+	# 		print(name)
+	# 		total_sum = total_sum + donor_obj.sum_donation
+	# 	return total_sum
+
+	def get_donor(self, donor_name):
+	 	return self._donors[donor_name.lower()]
+
+	def print_Db(self):
+		for name,donor_obj in self._donors.items(): 
+			print("\n{} donated individual donations of ${} for a total of ${}".format(name,donor_obj.donations,donor_obj.sum_donation))
```
PythonCHB, line 58: ideally this would return a report a a string, to printed elsewhere -- more flexible, and easier to test.

students/aravichander/lesson09/Mailroom_class.py
```diff
@@ -3,68 +3,105 @@
...
+
+	def get_sorted_donors(self): 
+		return sorted(self._donors)
```
PythonCHB, line 61: nice!

students/aravichander/lesson09/Mailroom_class.py
```diff
@@ -3,68 +3,105 @@
...
+	donor_db.add_donor(e)
+	return donor_db
+
+def thankall(db):
+	for name in db.get_sorted_donors():
+		donor = db.get_donor(name)
+		totaldonated = donor.sum_donation
+		letterofthanks(name,totaldonated)
+
+def letterofthanks(donorname,totaldonated): #Potentially create method for letterofthanks within donor class
```
PythonCHB, line 104: if you make `letter_of_thanks()` a method of the Donor class, then it would have all the information it needed (in self)

students/aravichander/lesson09/Mailroom_class.py
```diff
@@ -73,54 +110,57 @@ def letterofthanks(donorname,totaldonated):
...
-def searchdonors(thankdonor):
+def searchdonors(thankdonor,db):
 		try:
-			print(Donors[thankdonor])
-			totaldonated = sumdonor(thankdonor)
+			#print(db[thankdonor])
+			donor = db.get_donor(thankdonor)
+			totaldonated = donor.sum_donation
 			letterofthanks(thankdonor,totaldonated)
 		except KeyError:
```
PythonCHB, line 122: nice use of exception handling!

students/aravichander/lesson09/Mailroom_class.py
```diff
@@ -73,54 +110,57 @@ def letterofthanks(donorname,totaldonated):
...
-def newdonation():
+def newdonation(db):
 	newdonor = input("Input name of new donor\n")
 	while len(str.strip(newdonor)) == 0:
 		newdonor = input("Invalid input, enter name of new donor\n")
 	newdonor = str.strip(newdonor)
-	Donors[newdonor] = []
+	try:
+		newdonor_object = db.get_donor(newdonor)
+	except:
```
PythonCHB, line 133: never use  bare Except!

this would be a KeyError, yes?

students/aravichander/lesson09/Mailroom_class.py
```diff
@@ -73,54 +110,57 @@ def letterofthanks(donorname,totaldonated):
...
+	db.print_Db()
 
-def thanks():
+def thanks(db):
 	thankdonor = input("\nPlease enter full name of donor you'd like to thank or list to view donors\n")
 	if thankdonor == "list":
-		printdonors()
-		thanks()
+		db.print_Db()
+		thanks(db)
```
PythonCHB, line 153: careful with recursion in an interactive loop -- this should be a while loop, so it can go on forever if need be.

students/aravichander/lesson09/test_mailroom.py
```diff
@@ -0,0 +1,34 @@
+
+import unittest
+from Mailroom_class import Donor, Donor_db
+
+class MyFuncTestCase(unittest.TestCase):
```
PythonCHB, line 5: you want to use meaningful names for your test classes.

students/aravichander/lesson09/test_mailroom.py
```diff
@@ -0,0 +1,34 @@
...
+	def test_donor_class(self):
```
PythonCHB, line 6: this is putting way too much in a single test.

each method / property should get its own test.

when they all pass, this is fine, but if one fails, the rest of the tests won't be run -- and you want all of them to run every time.

students/aravichander/lesson09/test_mailroom.py
```diff
@@ -0,0 +1,34 @@
...
+		self.assertEqual(b.number_of_donations,3)
+		self.assertEqual(b.donation_average,600)
+
+	def test_donor_db_class(self):
+		a = Donor("Foghorn Leghorn",[100,200])
+		b = Donor("Bugs Bunny",[500,600])
+		donor_db = Donor_db()
+		donor_db.add_donor(a)
+		donor_db.add_donor(b)
+		self.assertEqual(donor_db.get_sorted_donors(),['bugs bunny','foghorn leghorn'])
```
PythonCHB, line 28: this should also be three separate tests.

PythonCHB (commented):
you need to do a bit more separation of logic from interaction and input/output, but you've got a good start here.

And you need to split your tests out -- each test should test one thing -- it may have a couple asserts, but shouldn't have more than a few, and they should be closely related.

PythonCHB:
I gave you credit for mailroom testing, and OO mailroom.

Congratulations! You've reached 84% complete.

But I highly encourage you to do the html render assignment, and go in and cleanup OO mailroom some more -- you need to gain more comfort with classes, OO programming and testing.

